PR #29, opened 2016-01-03
Fix bug in basic_iterator operator--(int). Add test case for same.

Beman:
Added test case was failing before fix. OK after fix.

HTH,

--Beman

### DIFF
--- a/include/stl2/detail/iterator/basic_iterator.hpp
+++ b/include/stl2/detail/iterator/basic_iterator.hpp
@@ -722,7 +722,7 @@ STL2_OPEN_NAMESPACE {
     {
       auto tmp = *this;
       --*this;
-      return *this;
+      return tmp;
     }
 
     constexpr basic_iterator& operator+=(difference_type n) &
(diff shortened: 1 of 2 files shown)


include/stl2/detail/iterator/basic_iterator.hpp
```diff
@@ -722,7 +722,7 @@ STL2_OPEN_NAMESPACE {
     {
       auto tmp = *this;
       --*this;
-      return *this;
+      return tmp;
```
ericniebler: Or simply `return prev(*this);`

CaseyCarter: > Or simply `return prev(*this);`

We want the "old" value which is `next(*this)`, not `prev(*this)`. `--*this; return next(*this);` would be semantically equivalent, but I conjecture that `auto tmp = *this; --*this; return tmp;` is cheaper in general. A copy must be made in any case, copying the original value will be cheaper if the optimizer can't determine that `++` simply reverses the effect of `--`.

ericniebler: Sorry, looked

ericniebler: too quickly. :-P

PR #43, opened 2016-10-27
Implement PR for ericniebler/stl2#239:

CaseyCarter:
In `indirect_result_of`, decay the type of the function before passing to `IndirectCallable`.


ericniebler:
LGTM

PR #44, opened 2016-10-27
Nuke callable wrappers

ericniebler:


CaseyCarter:
Travis is busy chewing on something else, but this works on my machine.

PR #47, opened 2016-10-28
update algorithms for perfectly forwarding Callable concept

ericniebler:
refs ericniebler/stl2#189

### DIFF
--- a/include/stl2/detail/algorithm/generate.hpp
+++ b/include/stl2/detail/algorithm/generate.hpp
@@ -20,7 +20,10 @@
 // generate [alg.generate]
 //
 STL2_OPEN_NAMESPACE {
-	template <Callable F, OutputIterator<result_of_t<F&()>> O, Sentinel<O> S>
+	template <class F, Iterator O, Sentinel<O> S>
+	requires
+		models::Callable<F&> &&
+		models::Writable<O, result_of_t<F&()>>
 	O generate(O first, S last, F gen)
 	{
 		for (; first != last; ++first) {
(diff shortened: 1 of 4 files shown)


include/stl2/detail/algorithm/generate.hpp
```diff
@@ -20,7 +20,9 @@
 // generate [alg.generate]
 //
 STL2_OPEN_NAMESPACE {
-	template <Callable F, OutputIterator<result_of_t<F&()>> O, Sentinel<O> S>
+	template <class F, OutputIterator<result_of_t<F&()>> O, Sentinel<O> S>
+	requires
+		models::Callable<F&>
```
CaseyCarter: It's too late to require `Callable<F&>`, we've already instantiated `result_of_t<F&()>`. This needs to be:
```c++
template <class F, Iterator O, Sentinel<O> S>
requires
    models::Callable<F&> &&
    models::Writable<O, result_of_t<F&()>>
```

include/stl2/detail/algorithm/generate_n.hpp
```diff
@@ -20,7 +20,9 @@
 // generate_n [alg.generate]
 //
 STL2_OPEN_NAMESPACE {
-	template <Callable F, OutputIterator<result_of_t<F&()>> O>
+	template <class F, OutputIterator<result_of_t<F&()>> O>
+	requires
+		models::Callable<F&>
```
CaseyCarter: Same issue here as for `generate`.

ericniebler:
Better?

CaseyCarter:
Yip.

PR #49, opened 2016-11-02
Memory

cjdb:
Pre-renaming of `__ReferenceTo`.
Tests are also unique at the moment, I'll write some tests that use `CHECK` soon.
What are your initial thoughts?


test/memory/uninitialized_move.cpp
```diff
@@ -0,0 +1,107 @@
...
+   }
+};
+
+template <typename T>
+struct Assert<T, true>
+{
+   static void empty(const Array<T>&)
+   {
+   }
+};
```
CaseyCarter: Concept dispatch would make this a little cleaner:
```c++
template <typename T>
void assert_empty(const Array<T>& a)
{
    /* ... */
}

template <typename T>
requires ranges::is_fundamental<T>::value
void assert_empty(const Array<T>&)
{}
```

test/memory/uninitialized_move.cpp
```diff
@@ -0,0 +1,107 @@
...
+   uninitialized_move_test(Test_type_one{0, 1, 2, 3, 4, 5, 6, 7});
+   uninitialized_move_test(Test_type_two{{{0.0, 0.1, 0.2},
+                                          {1.0, 1.1, 1.2, 1.3, 1.4},
+                                          {2.0, 2.1, 2.2, 2.3},
+                                          {3.01, 3.20, 3.33, 3.4},
+                                          {4.101, 4.102, 4.201, 4.202, 4.311},
+                                          {5.},
+                                          {6.1, 3.02, 6.3, 6.4, 6.5, 6.6, 6.7, 6.8, 6.9},
+                                          std::vector<double>(1 << 26, 7.0)}});
+}
```
CaseyCarter: I find it odd that `uninitialized_move` is never tested with a move-only type.

include/stl2/detail/memory/destroy.hpp
```diff
@@ -0,0 +1,76 @@
...
+   // destroy [Extension]
+   //
+   template <InputIterator I, Sentinel<I> S>
+   requires
+      Destructible<value_type_t<I>>() &&
+      __ReferenceTo<I, value_type_t<I>>()
+   I destroy(I first, S last) noexcept
+   {
+      for (; first != last; ++first)
+         destroy_at(__stl2::addressof(*first));
```
cjdb: ADL is the bane of my C++ programming: I know it exists, I generally know how it works, but can never remember the specifics. It's high time that I learn its rules properly.

include/stl2/detail/memory/uninitialized_copy.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <stl2/detail/fwd.hpp>
+#include <stl2/detail/memory/destroy.hpp>
+#include <stl2/detail/memory/reference_to.hpp>
+#include <stl2/detail/tagged.hpp>
+
+STL2_OPEN_NAMESPACE {
+   ///////////////////////////////////////////////////////////////////////////
+   // uninitialized_copy [Extension]
+   //
+   template <InputIterator I, Sentinel<I> S, /*NoThrow*/ForwardIterator O>
```
cjdb: Ahh, so it's a semantic difference! I think I got hung up on `Destructible` having

```cpp
{ t.~T() } noexcept;
```

and thought the same could be applied to `NoThrowForwardIterator`. I'll fix in `stl2/proposals-master` as well. The sooner we can get axioms, the better (toying with making researching into it a PhD topic if I can come up with a cool enough idea and can brave the four years -- no promises though).

cjdb: Should there be an exposition `__NoThrowForwardRange`?

```cpp
template <typename T>
concept bool __NoThrowForwardRange {
   return ForwardRange<T>();
   // Axiom: as with __NoThrowForwardIterator
}
```

include/stl2/detail/memory/uninitialized_copy.hpp
```diff
@@ -0,0 +1,72 @@
...
+      Constructible<value_type_t<O>, reference_t<I>>() &&
+      __ReferenceTo<O, value_type_t<O>>()
+   tagged_pair<tag::in(I), tag::out(O)>
+   uninitialized_copy(I first, S last, O result)
+   {
+      auto saved = result;
+      try {
+         for (; first != last; ++result, (void)++first) {
+            ::new(const_cast<void*>(static_cast<const volatile void*>(__stl2::addressof(*result))))
+               value_type_t<I>(*first);
```
cjdb: > `s/value_type_t<I>/value_type_t<O>/`

Proof my tests are not exhaustive :)

> Actually, I think all of the `::new`s would be better replaced with a helper function (somewhere under ) that deduces the proper type...
...so we don't have to keep reading that poor tortured placement new over and over.

We're standardising `destroy_at`; why not standardise `construct_at` too?

cjdb: Admittedly, that would mean we'd need:

```cpp
template <typename T, typename... Args>
  requires Constructible<T, Args...>()
void construct_at(T& t, Args&&... args);

template <typename T>
   requires DefaultConstructible<T>()
void default_construct_at(T& t);

template <typename T>
   requires DefaultConstructible<T>()
void value_construct_at(T& t);
```

and potentially pointer-based ones as well (to maintain consistency with `destroy_at`).

include/stl2/detail/memory/uninitialized_copy.hpp
CaseyCarter: > Ahh, so it's a semantic difference!

Yes. We could feasibly *require* the operations to be `noexcept` with Concepts, but it would be a breaking change with Standard C++, and not a hugely valuable break at that. We decided that actually requiring `noexcept` for `Destructible` was not too breaking given that destructors are implicitly `noexcept` for most sane types.

> Should there be an exposition `__NoThrowForwardRange`?
`__NoThrowForwardRange<T>` is `Range<T>() && __NoThrowForwardIterator<iterator_t<T>>()`. If you prefer to type `__NoThrowForwardRange<T>`, then by all means define it. I have no preference.

include/stl2/detail/memory/uninitialized_copy.hpp
CaseyCarter: > why not standardise `construct_at` too?

That would be better in a separate tiny proposal targeted at the C++ WP. I don't know that it provides enough value; maybe ask the SG14 folks for their thoughts?

cjdb: Does this mean that we _can't_ sneak that `destroy` overload into the memory proposal?

cjdb:
Investigating check failure, could have been a failure to rebuild before testing.

cjdb:
Nope, still passes after `make clean && make -j 16 && make test`. Switching to manual testing.

CaseyCarter:
Merged this along with a few cleanup commits in a5988a9d2f06dbe5d1bdcdb66a54dc9f601f149c. Thanks.

PR #52, opened 2016-11-08
add any_input_iterator

### LINKED ISSUE #51: type-erased any_input_iterator
I'll just leave this here. Caveat: largely untested and probably buggy as hell.

```c++
//  Copyright Eric Niebler 2016
//
//  Use, modification and distribution is subject to the
//  Boost Software License, Version 1.0. (See accompanying
//  file LICENSE_1_0.txt or copy at
//  http://www.boost.org/LICENSE_1_0.txt)
namespace any_iterator_detail
{
	enum class op { copy, move, nuke, bump, comp };	

	struct counted
	{
		std::atomic<long> cnt{ 1 };
	};
	ranges::InputIterator{I}
	struct shared_iterator : counted
	{
		shared_iterator(I i) : it(std::move(i)) {}
		I it;
	};

	union blob
	{
		counted* big;
		std::aligned_storage<2 * sizeof(void*)>::type tiny;	
	};	

	template <typename It>	
	using is_small =
	    std::integral_constant<bool, (sizeof(It) <= sizeof(blob::tiny))>;
	using small_tag = std::true_type;	
	using big_tag = std::false_type;

	inline bool uninit_noop(op, blob*, blob*)
	{
		return false;	
	}

	template<typename Reference>
	[[noreturn]] inline Reference uninit_deref(blob const &)
	{
		std::terminate();
	}	

	template <typename Reference, ranges::InputIterator I>
	static Reference deref_small(blob const & p)
	{
		return **static_cast<I const *>(static_cast<void const *>(&p.tiny));
	}
		
	template <typename Reference, ranges::InputIterator I>
	static Reference deref_big(blob const& p)
	{	
		return *static_cast<shared_iterator<I> const *>(p.big)->it;
	}	

	ranges::InputIterator{I}	
	bool exec_small(op o, blob* src, blob* dst)
	{	
		switch (o)
		{
		case op::copy:
			::new (static_cast<void*>(&dst->tiny))
				I(*static_cast<I const *>(static_cast<void*>(&src->tiny)));
			break;
		case op::move:
			::new (static_cast<void*>(&dst->tiny))
				I(std::move(*static_cast<I*>(static_cast<void*>(&src->tiny))));	
			// fallthrough
		case op::nuke:
			static_cast<I*>(static_cast<void*>(&src->tiny))->~I();	
			break;	
		case op::bump:
			++*static_cast<I*>(static_cast<void*>(&src->tiny));
			break;
		case op::comp:
			return *static_cast<I const *>(static_cast<void*>(&src->tiny))
			    == *static_cast<I const *>(static_cast<void*>(&dst->tiny));
		}	
		return false;	
	}	
		
	ranges::InputIterator{I}	
	bool exec_big(op o, blob* src, blob* dst)
	{
		switch (o)
		{
		case op::copy:
			++(dst->big = src->big)->cnt;
			break;
		case op::move:
			dst->big = std::exchange(src->big, nullptr);
			break;
		case op::nuke:
			if (0 == --src->big->cnt)
				delete static_cast<shared_iterator<I>*>(src->big);
			break;
		case op::bump:
			++static_cast<shared_iterator<I>*>(src->big)->it;
			break;
		case op::comp:
			return static_cast<shared_iterator<I> const *>(src->big)->it
			    == static_cast<shared_iterator<I> const *>(dst->big)->it;
		}
		return true;	
	}	

	template<class Reference, class ValueType>
	struct any_input_cursor
	{
	private:
		blob data_ = { nullptr };
		Reference (*deref_)(blob const &) = &uninit_deref<Reference>;
		bool (*exec_)(op, blob*, blob*) = &uninit_noop;

		ranges::InputIterator{I} any_input_cursor(I i, small_tag)
		{
			::new (static_cast<void*>(&data_.tiny)) I(std::move(i));
			deref_ = &deref_small<Reference, I>;
			exec_ = &exec_small<I>;
		}
		ranges::InputIterator{I} any_input_cursor(I i, big_tag)
		{
			data_.big = new shared_iterator<I>(std::move(i));
			deref_ = &deref_big<Reference, I>;
			exec_ = &exec_big<I>;
		}
	public:
		using value_type = ValueType;
		using single_pass = std::true_type;
		any_input_cursor() = default;
		ranges::InputIterator{I} any_input_cursor(I i)
			requires ranges::ConvertibleTo<ranges::reference_t<I>, Reference>()
		: any_input_cursor(std::move(i), is_small<I>{}) {}
		any_input_cursor(any_input_cursor &&that)
		{
			that.exec_(op::move, &that.data_, &data_);
			ranges::swap(deref_, that.deref_);
			ranges::swap(exec_, that.exec_);
		}
		any_input_cursor(any_input_cursor const &that)
		{
			that.exec_(op::copy, const_cast<blob*>(&that.data_), &data_);
			deref_ = that.deref_;
			exec_ = that.exec_;
		}
		any_input_cursor& operator=(any_input_cursor &&that)
		{
			if (&that != this)
			{
				this->~any_input_cursor();
				::new (static_cast<void*>(this))
					any_input_cursor(std::move(that));
			}
			return *this;
		}
		any_input_cursor& operator=(any_input_cursor const &that)
		{
			if (&that != this)
			{
				this->~any_input_cursor();
				::new (static_cast<void*>(this)) any_input_cursor(that);
			}
			return *this;
		}
		~any_input_cursor()
		{
			exec_(op::nuke, &data_, nullptr);
		}
		Reference read() const
		{
			return deref_(data_);
		}
		bool equal(any_input_cursor const &that) const
		{
			return exec_(op::comp, const_cast<blob*>(&data_),
				const_cast<blob*>(&that.data_));
		}
		void next()
		{
			exec_(op::bump, &data_, nullptr);
		}
	};
}

template<class Reference, class ValueType =
	std::remove_cv_t<std::remove_reference_t<Reference>>>
using any_input_iterator =
	ranges::basic_iterator<
		any_iterator_detail::any_input_cursor<Reference, ValueType>>;
```

ericniebler:
'Cause why not? Does the small object optimization. Fixes #51

### DIFF
--- a/include/stl2/detail/iterator/any_iterator.hpp
+++ b/include/stl2/detail/iterator/any_iterator.hpp
@@ -1,0 +1,230 @@
+// cmcstl2 - A concept-enabled C++ standard library
+//
+//  Copyright Eric Niebler 2016
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/caseycarter/cmcstl2
+//
+#ifndef STL2_DETAIL_ITERATOR_ANY_ITERATOR_HPP
+#define STL2_DETAIL_ITERATOR_ANY_ITERATOR_HPP
+
+#include <new>
+#include <atomic>
+#include <stl2/type_traits.hpp>
+#include <stl2/detail/fwd.hpp>
+#include <stl2/detail/swap.hpp>
+#include <stl2/detail/concepts/core.hpp>
+#include <stl2/detail/iterator/basic_iterator.hpp>
+#include <stl2/detail/iterator/concepts.hpp>
+
+STL2_OPEN_NAMESPACE {
+	namespace __any_iterator {
+        enum class op { copy, move, nuke, bump, comp, rval };
+
+        struct counted {
+            std::atomic<long> cnt{ 1 };
+        };
+        InputIterator{I} struct shared_iterator : counted {
+            shared_iterator(I i) : it(std::move(i)) {}
+            I it;
+        };
+
+        union blob {
+            counted* big;
+            std::aligned_storage_t<2 * sizeof(void *)> tiny;
+        };
+
+        template <class It>
+        using is_small =
+            std::integral_constant<bool, (sizeof(It) <= sizeof(blob::tiny))>;
+        using small_tag = std::true_type;
+        using big_tag = std::false_type;
+
+        template <class RValueReference>
+        using iter_move_fn = RValueReference (*)(blob const &);
+
+        template <class RValueReference>
+        inline iter_move_fn<RValueReference> uninit_noop(op, blob *, blob *) {
+            return nullptr;
+        }
+
+        template <class Reference>
+        [[noreturn]] inline Reference uninit_deref(blob const &) {
+            std::terminate();
+        }
+
+        template <class Reference, InputIterator I>
+        static Reference deref_small(blob const &src) {
+            return **static_cast<I const *>(static_cast<void const *>(&src.tiny));
+        }
+
+        template <class Reference, InputIterator I>
+        static Reference deref_big(blob const &src) {
+            return *static_cast<shared_iterator<I> const *>(src.big)->it;
+        }
+
+        template <class RValueReference, InputIterator I>
+        iter_move_fn<RValueReference> exec_small(op o, blob *src, blob *dst) {
+            switch (o) {
+            case op::copy:
+                ::new (static_cast<void *>(&dst->tiny))
+                    I(*static_cast<I const *>(static_cast<void *>(&src->tiny)));
+                break;
+            case op::move:
+                ::new (static_cast<void *>(&dst->tiny))
+                    I(std::move(*static_cast<I *>(static_cast<void *>(&src->tiny))));
+                // fallthrough
+            case op::nuke:
+                static_cast<I *>(static_cast<void *>(&src->tiny))->~I();
+                break;
+            case op::bump:
+                ++*static_cast<I *>(static_cast<void *>(&src->tiny));
+                break;
+            case op::comp:
+                if (*static_cast<I const *>(static_cast<void *>(&src->tiny)) ==
+                    *static_cast<I const *>(static_cast<void *>(&dst->tiny))) {
+                    // fallthrough
+            case op::rval:
+                    return +[](blob const &src) -> RValueReference {
+                        return iter_move(
+                            *static_cast<I const *>(static_cast<void const *>(&src.tiny)));
+                    };
+                }
+            }
+            return nullptr;
+        }
+
+        template <class RValueReference, InputIterator I>
+        iter_move_fn<RValueReference> exec_big(op o, blob *src, blob *dst) {
+            switch (o) {
+            case op::copy:
+                ++(dst->big = src->big)->cnt;
+               break;
+            case op::move:
+                dst->big = std::exchange(src->big, nullptr);
+                break;
+            case op::nuke:
+                if (0 == --src->big->cnt)
+                    delete static_cast<shared_iterator<I>*>(src->big);
+                break;
+            case op::bump:
+                ++static_cast<shared_iterator<I> *>(src->big)->it;
+                break;
+            case op::comp:
+                if (static_cast<shared_iterator<I> const *>(src->big)->it ==
+                    static_cast<shared_iterator<I> const *>(dst->big)->it) {
+                    // fallthrough
+            case op::rval:
+                    return +[](blob const &src) -> RValueReference {
+                        return iter_move(
+                            static_cast<shared_iterator<I> const *>(src.big)->it);
+                    };
+                }
+            }
+            return nullptr;
+        }
+
+        template<class Reference, class ValueType, class RValueReference>
+        struct cursor {
+        private:
+            blob data_ = { nullptr };
+            Reference (*deref_)(blob const &) =
+                &uninit_deref<Reference>;
+            iter_move_fn<RValueReference> (*exec_)(op, blob *, blob *) =
+                &uninit_noop<RValueReference>;
+
+            InputIterator{I} cursor(I i, small_tag) {
+                ::new (static_cast<void *>(&data_.tiny)) I(std::move(i));
+                deref_ = &deref_small<Reference, I>;
+                exec_ = &exec_small<RValueReference, I>;
+            }
+            InputIterator{I} cursor(I i, big_tag) {
+                data_.big = new shared_iterator<I>(std::move(i));
+                deref_ = &deref_big<Reference, I>;
+                exec_ = &exec_big<RValueReference, I>;
+            }
+        public:
+            using value_type = ValueType;
+            using single_pass = std::true_type;
+            struct mixin;
+            cursor() = default;
+            cursor(cursor &&that) {
+                that.exec_(op::move, &that.data_, &data_);
+                swap(deref_, that.deref_);
+                swap(exec_, that.exec_);
+            }
+            cursor(cursor const &that) {
+                that.exec_(op::copy, const_cast<blob *>(&that.data_), &data_);
+                deref_ = that.deref_;
+                exec_ = that.exec_;
+            }
+            cursor &operator=(cursor &&that) {
+                if (&that != this) {
+                    this->~cursor();
+                    ::new (static_cast<void *>(this))
+                        cursor(std::move(that));
+                }
+                return *this;
+            }
+            cursor &operator=(cursor const &that) {
+                if (&that != this) {
+                    this->~cursor();
+                    ::new (static_cast<void *>(this)) cursor(that);
+                }
+                return *this;
+            }
+            ~cursor() {
+                exec_(op::nuke, &data_, nullptr);
+            }
+            Reference read() const {
+                return deref_(data_);
+            }
+            bool equal(cursor const &that) const {
+                return exec_(op::comp, const_cast<blob *>(&data_),
+                    const_cast<blob *>(&that.data_)) != nullptr;
+            }
+            void next() {
+                exec_(op::bump, &data_, nullptr);
+            }
+            RValueReference indirect_move() const {
+                return exec_(op::rval, nullptr, nullptr)(data_);
+            }
+        };
+
+        template<class Reference, class ValueType, class RValueReference>
+        struct cursor<Reference, ValueType, RValueReference>::mixin
+        : basic_mixin<cursor> {
+            mixin() = default;
+            using mixin::basic_mixin::basic_mixin;
+            InputIterator{I} mixin(I i)
+            requires
+                ConvertibleTo<reference_t<I>, Reference>() &&
+                ConvertibleTo<rvalue_reference_t<I>, RValueReference>()
+            : basic_mixin<cursor>(cursor{std::move(i), is_small<I>{}}) {
+            }
+            template <class I, class C = cursor,
+                class = meta::if_c<!is_same<I, mixin>::value>>
+            requires
+                !Same<I, basic_iterator<C>>() && InputIterator<I>() &&
+                ConvertibleTo<reference_t<I>, Reference>() &&
+                ConvertibleTo<rvalue_reference_t<I>, RValueReference>()
+            mixin &operator=(I i) & {
+                this->get() = cursor{i, is_small<I>{}};
+                return *this;
+            }
+        };
+	}
+
+    template<class Reference,
+        class ValueType = __uncvref<Reference>,
+        class RValueReference = __iter_move::rvalue<Reference>>
+    using any_input_iterator =
+        basic_iterator<__any_iterator::cursor<Reference, ValueType, RValueReference>>;
+
+} STL2_CLOSE_NAMESPACE
+
+#endif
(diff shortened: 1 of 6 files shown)


include/stl2/detail/iterator/any_iterator.hpp
```diff
@@ -0,0 +1,230 @@
...
+            InputIterator{I} mixin(I i)
+            requires
+                ConvertibleTo<reference_t<I>, Reference>() &&
+                ConvertibleTo<rvalue_reference_t<I>, RValueReference>()
+            : basic_mixin<cursor>(cursor{std::move(i), is_small<I>{}}) {
+            }
+            template <class I, class C = cursor,
+                class = meta::if_c<!is_same<I, mixin>::value>>
+            requires
+                !Same<I, basic_iterator<C>>() && InputIterator<I>() &&
```
ericniebler: It would be interesting to know why I needed to use SFINAE here (`meta::if_c<!is_same...>`) instead of concepts (`!Same<...`). Also interesting about this: I couldn't form `basic_iterator<cursor>` here because `basic_iterator`'s template parameter is required to satisfy `Cursor`, and concept checking gets into some weird recursive instantiation SNAFU.

include/stl2/detail/iterator/any_iterator.hpp
```diff
@@ -0,0 +1,230 @@
...
+            std::atomic<long> cnt{ 1 };
+        };
+        InputIterator{I} struct shared_iterator : counted {
+            shared_iterator(I i) : it(std::move(i)) {}
+            I it;
+        };
+
+        union blob {
+            counted* big;
+            std::aligned_storage_t<2 * sizeof(void *)> tiny;
```
ericniebler: Iterators smaller than or equal to two pointers are stored in-situ. The size of `any_input_iterator` overall is 4 pointers. We can argue if these are the right thresholds.

include/stl2/detail/iterator/any_iterator.hpp
```diff
@@ -0,0 +1,230 @@
...
+            cursor(cursor const &that) {
+                that.exec_(op::copy, const_cast<blob *>(&that.data_), &data_);
+                deref_ = that.deref_;
+                exec_ = that.exec_;
+            }
+            cursor &operator=(cursor &&that) {
+                if (&that != this) {
+                    this->~cursor();
+                    ::new (static_cast<void *>(this))
+                        cursor(std::move(that));
```
CaseyCarter: destroy+construct is no bueno when the construction can throw. (and on 176)

include/stl2/detail/iterator/any_iterator.hpp
```diff
@@ -0,0 +1,230 @@
...
+                    I(std::move(*static_cast<I *>(static_cast<void *>(&src->tiny))));
+                // fallthrough
+            case op::nuke:
+                static_cast<I *>(static_cast<void *>(&src->tiny))->~I();
+                break;
+            case op::bump:
+                ++*static_cast<I *>(static_cast<void *>(&src->tiny));
+                break;
+            case op::comp:
+                if (*static_cast<I const *>(static_cast<void *>(&src->tiny)) ==
```
CaseyCarter: ... or `==` could always return true when `I` doesn't satisfy `Sentinel<I, I>()`; though that would be easier implemented with a concrete vtable than this vtable-in-a-function.

include/stl2/detail/iterator/any_iterator.hpp
```diff
@@ -0,0 +1,230 @@
...
+            return nullptr;
+        }
+
+        template <class Reference>
+        [[noreturn]] inline Reference uninit_deref(blob const &) {
+            std::terminate();
+        }
+
+        template <class Reference, InputIterator I>
+        static Reference deref_small(blob const &src) {
```
CaseyCarter: Why `static` here and on 66? Are these ex-member functions?

include/stl2/detail/iterator/any_iterator.hpp
```diff
@@ -0,0 +1,230 @@
...
+                ConvertibleTo<rvalue_reference_t<I>, RValueReference>()
+            : basic_mixin<cursor>(cursor{std::move(i), is_small<I>{}}) {
+            }
+            template <class I, class C = cursor,
+                class = meta::if_c<!is_same<I, mixin>::value>>
+            requires
+                !Same<I, basic_iterator<C>>() && InputIterator<I>() &&
+                ConvertibleTo<reference_t<I>, Reference>() &&
+                ConvertibleTo<rvalue_reference_t<I>, RValueReference>()
+            mixin &operator=(I i) & {
```
CaseyCarter: IIRC, `basic_iterator` provides this assignment for types that convert to `cursor`. Why not make `cursor` implicitly constructible from `I`?

CaseyCarter: Correction: when `bar` converts to `cursor`, `i = bar` is equivalent to `i = basic_iterator<cursor>{bar}`.

include/stl2/detail/iterator/any_iterator.hpp
ericniebler: Yes, a hold over from an earlier design.

include/stl2/detail/iterator/any_iterator.hpp
```diff
@@ -0,0 +1,230 @@
...
+            }
+        public:
+            using value_type = ValueType;
+            using single_pass = std::true_type;
+            struct mixin;
+            cursor() = default;
+            cursor(cursor &&that) {
+                that.exec_(op::move, &that.data_, &data_);
+                swap(deref_, that.deref_);
+                swap(exec_, that.exec_);
```
CaseyCarter: These are swapping uninitialized members with the members of `that.`.

include/stl2/detail/iterator/any_iterator.hpp
ericniebler: I think I can manage.

include/stl2/detail/iterator/any_iterator.hpp
ericniebler: Sí, but I'm more concerned about performance, and I worry about small iterators that are no-throw in fact but not marked `noexcept`.

include/stl2/detail/iterator/any_iterator.hpp
CaseyCarter: I missed the default-member-initializers; nevermind.

include/stl2/detail/iterator/any_iterator.hpp
CaseyCarter: More concerned about performance than correctness? I would rather mark these functions `noexcept` so the program terminates if the construction throws than enjoy the undefined behavior that will result from double destruction of `*this`.

ericniebler: If you are still concerned, I can move iterators that aren't nothrow move-constructible (ick) to the heap.

ericniebler: The UB is harmless in this case. You'd end up with an empty `any_input_iterator` that has nothing in the buffer and pointers to functions that are no-ops.

CaseyCarter: I think a sufficiently smart compiler can break this, and even more I detest trying to reason about UB. I'd prefer a formulation without the defined behavior a la:

```diff
@@ -147,33 +147,45 @@ STL2_OPEN_NAMESPACE {
                 deref_ = &deref_big<Reference, I>;
                 exec_ = &exec_big<RValueReference, I>;
             }
+            void reset() noexcept {
+                exec_(op::nuke, &data_, nullptr);
+                deref_ = &uninit_deref<Reference>;
+                exec_ = &uninit_noop<RValueReference>;
+            }
+            void copy_from(cursor const &that) {
+                // Pre: *this is empty
+                that.exec_(op::copy, const_cast<blob *>(&that.data_), &data_);
+                deref_ = that.deref_;
+                exec_ = that.exec_;
+            }
+            void move_from(cursor &that) {
+                // Pre: *this is empty
+                that.exec_(op::move, &that.data_, &data_);
+                swap(deref_, that.deref_);
+                swap(exec_, that.exec_);
+            }
         public:
             using value_type = ValueType;
             using single_pass = std::true_type;
             struct mixin;
             cursor() = default;
             cursor(cursor &&that) {
-                that.exec_(op::move, &that.data_, &data_);
-                swap(deref_, that.deref_);
-                swap(exec_, that.exec_);
+                move_from(that);
             }
             cursor(cursor const &that) {
-                that.exec_(op::copy, const_cast<blob *>(&that.data_), &data_);
-                deref_ = that.deref_;
-                exec_ = that.exec_;
+                copy_from(that);
             }
             cursor &operator=(cursor &&that) {
                 if (&that != this) {
-                    this->~cursor();
-                    ::new (static_cast<void *>(this))
-                        cursor(std::move(that));
+                    reset();
+                    move_from(that);
                 }
                 return *this;
             }
             cursor &operator=(cursor const &that) {
                 if (&that != this) {
-                    this->~cursor();
-                    ::new (static_cast<void *>(this)) cursor(that);
+                    reset();
+                    copy_from(that);
                 }
                 return *this;
             }
```

which seems to generate slightly *better* code from my brief investigation.

ericniebler: Picky, picky. ;-)

include/stl2/detail/iterator/any_iterator.hpp
CaseyCarter: I don't know that I *want* to support meaning for `i = bar` that would be different from `i = Iterator{bar}`; iterators are (generally) regular-ish. The simplest way to ensure that both syntaxes are supported with the same meaning is to use the same conversion constructor to implement them.

ericniebler: Regarding semantics, I agree. But it is potentially a perf win to support `op=` directly. Consider the difference between copy-constructing a container vs. copy-assigning a container. The latter has the chance to reuse storage. And some iterators (e.g. `istream_iterator<string>`) have members that are containers.

CaseyCarter: If the goal is only to support reasonable semantics, then I don't think we need a "protocol" hook for this in `mixin`; we should have an assignment operator overload in `basic_iterator` that simply forwards assignments to the cursor. This encourages slightly tighter semantics and at the same time makes it easier to assign to the cursor:

```diff
diff --git a/include/stl2/detail/iterator/any_iterator.hpp b/include/stl2/detail/iterator/any_iterator.hpp
index a4ae014..97f804a 100644
--- a/include/stl2/detail/iterator/any_iterator.hpp
+++ b/include/stl2/detail/iterator/any_iterator.hpp
@@ -159,6 +159,9 @@ STL2_OPEN_NAMESPACE {
                 deref_ = &deref_big<Reference, I>;
                 exec_ = &exec_big<RValueReference, I>;
             }
+            InputIterator{I} cursor(I i)
+            : cursor{i, is_small<I>{}}
+            {}
             void reset() noexcept {
                 exec_(op::nuke, &data_, nullptr);
                 deref_ = &uninit_deref<Reference>;
@@ -179,7 +182,6 @@ STL2_OPEN_NAMESPACE {
         public:
             using value_type = ValueType;
             using single_pass = std::true_type;
-            struct mixin;
             cursor() = default;
             cursor(cursor &&that) {
                 move_from(that);
@@ -218,30 +220,7 @@ STL2_OPEN_NAMESPACE {
                 return exec_(op::rval, nullptr, nullptr)(data_);
             }
         };
-
-        template<class Reference, class ValueType, class RValueReference>
-        struct cursor<Reference, ValueType, RValueReference>::mixin
-        : basic_mixin<cursor> {
-            mixin() = default;
-            using mixin::basic_mixin::basic_mixin;
-            InputIterator{I} mixin(I i)
-            requires
-                ConvertibleTo<reference_t<I>, Reference>() &&
-                ConvertibleTo<rvalue_reference_t<I>, RValueReference>()
-            : basic_mixin<cursor>(cursor{std::move(i), is_small<I>{}}) {
-            }
-            template <class I, class C = cursor,
-                class = meta::if_c<!is_same<I, mixin>::value>>
-            requires
-                !Same<I, basic_iterator<C>>() && InputIterator<I>() &&
-                ConvertibleTo<reference_t<I>, Reference>() &&
-                ConvertibleTo<rvalue_reference_t<I>, RValueReference>()
-            mixin &operator=(I i) & {
-                this->get() = cursor{i, is_small<I>{}};
-                return *this;
-            }
-        };
-	}
+    }
 
     template<class Reference,
         class ValueType = __uncvref<Reference>,
diff --git a/include/stl2/detail/iterator/basic_iterator.hpp b/include/stl2/detail/iterator/basic_iterator.hpp
index c16067d..13c25d8 100644
--- a/include/stl2/detail/iterator/basic_iterator.hpp
+++ b/include/stl2/detail/iterator/basic_iterator.hpp
@@ -602,12 +602,12 @@ STL2_OPEN_NAMESPACE {
 		// http://wg21.link/P0186
 		template <class O>
 		requires
-			__assignable<mixin&, O> && !models::Same<decay_t<O>, mixin> &&
-			!models::Same<decay_t<O>, basic_iterator>
+			!models::Same<decay_t<O>, basic_iterator> &&
+			models::Assignable<C&, O>
 		constexpr basic_iterator& operator=(O&& o) &
-		noexcept(is_nothrow_assignable<mixin&, O>::value)
+		noexcept(is_nothrow_assignable<C&, O>::value)
 		{
-			static_cast<mixin&>(*this) = forward<O>(o);
+			get() = __stl2::forward<O>(o);
 			return *this;
 		}
```

PR #55, opened 2016-11-26
Hack up move_view

CaseyCarter:


include/stl2/view/move.hpp
```diff
@@ -0,0 +1,83 @@
...
+#define STL2_DETAIL_VIEW_MOVE_HPP
+
+#include <stl2/detail/fwd.hpp>
+#include <stl2/detail/concepts/object.hpp>
+#include <stl2/detail/iterator/move_iterator.hpp>
+#include <stl2/detail/range/concepts.hpp>
+
+STL2_OPEN_NAMESPACE {
+	template <InputRange Rng>
+	class move_view {
```
viboes: Why name this class _view? I would think of it more like an adaptor, that change the behavior of the wrapped range.

include/stl2/view/move.hpp
```diff
@@ -0,0 +1,83 @@
...
+		move_sentinel<sentinel_t<Rng>> end()
+		{
+			return __stl2::make_move_sentinel(__stl2::end(rng_));
+		}
+		move_iterator<iterator_t<Rng>> end()
+		requires BoundedRange<Rng>()
+		{
+			return __stl2::make_move_iterator(__stl2::end(rng_));
+		}
+		auto size()
```
viboes: in which case do we need these non-const functions?

include/stl2/view/move.hpp
```diff
@@ -0,0 +1,83 @@
...
+//
+//  Use, modification and distribution is subject to the
+//  Boost Software License, Version 1.0. (See accompanying
+//  file LICENSE_1_0.txt or copy at
+//  http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/caseycarter/cmcstl2
+//
+#ifndef STL2_DETAIL_VIEW_MOVE_HPP
+#define STL2_DETAIL_VIEW_MOVE_HPP
```
CaseyCarter: Woops - my copypasta is showing.

include/stl2/view/move.hpp
CaseyCarter: In the parlance of range-v3, this is a "view adaptor". It adapts an underlying view into a different kind of view. range-v3 names all view adaptors `foo_view`, this is `move_view` for consistency with its range-v3 equivalent.

include/stl2/view/move.hpp
CaseyCarter: Fixed in #56.

include/stl2/view/move.hpp
CaseyCarter: Some views internally cache state that is mutated before or during iteration. The member functions which mutate that state are *physically* non-`const` despite that they do not *logically* modify the view by changing the sequence of elements. See ericniebler/range-v3#385 for extensive discussion. This design is widely disliked, but we haven't yet found anything better.

viboes: maybe it is worth having an issue here also that refers to the one in range-v3.

include/stl2/view/move.hpp
viboes: I dislike the name "view adaptor" as it is not a view, isn't it?
> "a view is essentially a pointer to a sequence of elements"

Maybe it s time to change this.

ericniebler: It is a view in that it is a sequence of elements owned by a different range. I believe view is the right term. It is crucial to getting people to think correctly about lifetime.

It is also an adaptor in that it lazily changes the underlying elements in some way.

viboes: Ok, I see. A range doesn't owns its elements, so a view on a range doesn't owns its elements neither. And view adaptor is a view a Range that in addition adapts the its contents.

I was confused as the `move_view` view adaptor copies the Range it is adapting, and so I didn't saw it as a view.

Sorry for the noise.

BTW, you meant owned by a different container, isn't it?

include/stl2/view/move.hpp
viboes: Ok, I'll take a deeper look on the issue and on the Range design.

include/stl2/view/move.hpp
CaseyCarter: A range may or may not own its contents. Views are the subset of ranges that are `Semiregular` and have copy and assignment that are O(1) relative to the number of elements. That O(1) requirement implies that views either do not own the elements they denote, or that they own a constant number of elements.

> I was confused as the move_view view adaptor copies the Range it is adapting, and so I didn't saw it as a view.

`InputRange` is not the correct constraint for the class parameter: it's required to satisfy both `View` and `InputRange` so that copy is guaranteed to be O(1). [That constraint is corrected in a later commit](https://github.com/CaseyCarter/cmcstl2/commit/49a0a43641f34ba1eb8170a1670a00ee3c0a8975#diff-5d0d31fea97eb927a7b79f10f1c47095L21).

PR #61, opened 2016-12-13
Extensions for Numeric

cjdb:
I have been working on functions in `<numeric>` for some time. Currently, these are the ones that I've implemented:

* `accumulate`
* `adjacent_difference`
* `inner_product`
* `partial_sum`
* `gcd`
* `lcm`

Early on, there was also work being put into a Standards proposal, but I have shelved that in favour of getting the implementations working: work on that shall resume after all C++17 numerical functions have been implemented.

Would it be possible to get some initial feedback on ranges/numeric please? There's still a long way to go, even for the functions that have been commit so far, but a few interesting cases have popped up, and I think it is time to touch base.

Please feel free to completely ignore `int128.hpp`.


cjdb:
The one downside to requiring a [`Regular`][`Strict`]`Number` is that it prevents using functions such as `adjacent_difference` on reasonable types such as `pair<int, int>`, which do not conform to `Number`, let alone the others.

Similarly, `partial_sum` on `string` is prohibited, since it too is not a `Number`.

Perhaps a `requires` clause such as

```cpp
template <class I, class O, class Op, class Proj = identity>
concept bool __AdjacentDifferentiable { // exposition only
   return RegularCallable<Op, 
}
```

would be better instead (similar to @ericniebler's `AdjacentDifferentiable`).

cjdb:
There seems to be an issue with `inner_product` taking a `vector<pair<int, int>>` and `vector<pair<int, double>>` when `Assignable<T&, __Result2>` (or similar) is required due to `CommonReference`.

I have contrasted this against [range-v3's `inner_product`](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/numeric/inner_product.hpp), which _does_ permit the above types.

See [lines 123 -- 131 of test/numeric/inner_product.cpp](https://github.com/cjdb/cmcstl2/blob/numerics/test/numeric/inner_product.cpp#L123) for the test in question.

ericniebler:
I'll note here that we have a US national body comment on the Ranges PDTS to add the numeric algorithms. That doesn't mean we have to provide them, just that we'll have to address the comment in Kona. Considering how tricky it is to get the concepts right for these algorithms, I'm inclined to put this off, but I haven't been tracking closely the work you guys have been doing here. Maybe the issues are settled?

cjdb:
> Considering how tricky it is to get the concepts right for these algorithms, I'm inclined to put this off

I'm in agreement, unless there's time to sink several hours into code reviews, including nutting out whatever is wrong with `inner_product` (I need help with getting the `Assignable` requirement right please).

cjdb:
See https://github.com/ericniebler/stl2/issues/229#issuecomment-275554891 for issues regarding `inner_product`.

cjdb:
FWIW, I do believe that the thread addresses and solves the issues I thought I was having. For C++14-related functions, only peer-reviewing and changes following reviews are necessary before moving forward with a proposal.

cjdb:
This is very out-of-date now. Something new to appear post-CppCon (hopefully pre-SD mailing deadline).

PR #64, opened 2017-01-18
experimental support for move_iterator post-fix increment

ericniebler:
Just experimenting to see how badly postfix increment is broken for generic iterator adaptors. (Motivated by my upcoming paper to kill Readability of op++ return type.) What do you think of this? Busted in some subtle way, or not?

### DIFF
--- a/include/stl2/detail/iterator/move_iterator.hpp
+++ b/include/stl2/detail/iterator/move_iterator.hpp
@@ -92,7 +92,7 @@ STL2_OPEN_NAMESPACE {
 			)
 
 			// Experimental support for move_iterator post-increment
-			using __post_inc_t = decltype(current_++);
+			using __postinc_t = std::decay_t<decltype(current_++)>;
 			Readable{R}
 			struct __proxy {
 				using value_type = __stl2::value_type_t<R>;
@@ -107,16 +107,11 @@ STL2_OPEN_NAMESPACE {
 				)
 			};
 
-			STL2_CONSTEXPR_EXT cursor post_increment()
-			noexcept(noexcept(cursor{cursor{__stl2::exchange(current_, __stl2::next(current_))}}))
-			requires Same<__post_inc_t, I>() {
-				return cursor{__stl2::exchange(current_, __stl2::next(current_))};
-			}
-
+			template <class = void> // BUGBUG why is this necessary?
 			STL2_CONSTEXPR_EXT auto post_increment()
-			noexcept(noexcept(current_++) && std::is_nothrow_move_constructible<__post_inc_t>::value)
-			requires !Same<__post_inc_t, I>() && Readable<__post_inc_t>() {
-				return __proxy<__post_inc_t>{current_++};
+			noexcept(noexcept(__proxy<__postinc_t>{current_++}))
+			requires !ForwardIterator<I>() && Readable<__postinc_t>() {
+				return __proxy<__postinc_t>{current_++};
 			}
 
 			STL2_CONSTEXPR_EXT void prev()
--- a/include/stl2/detail/iterator/move_iterator.hpp
+++ b/include/stl2/detail/iterator/move_iterator.hpp
@@ -18,7 +18,6 @@
 #include <stl2/detail/concepts/compare.hpp>
 #include <stl2/detail/concepts/object.hpp>
 #include <stl2/detail/iterator/basic_iterator.hpp>
-#include <stl2/detail/iterator/operations.hpp>
 #include <stl2/detail/iterator/concepts.hpp>
 
 STL2_OPEN_NAMESPACE {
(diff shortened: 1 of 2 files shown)


include/stl2/detail/iterator/move_iterator.hpp
```diff
@@ -90,6 +91,33 @@ STL2_OPEN_NAMESPACE {
...
+				friend decltype(auto) iter_move(const __proxy& that)
+				STL2_NOEXCEPT_RETURN(
+					__stl2::iter_move(that.__tmp)
+				)
+			};
+
+			STL2_CONSTEXPR_EXT cursor post_increment()
+			noexcept(noexcept(cursor{cursor{__stl2::exchange(current_, __stl2::next(current_))}}))
+			requires Same<__post_inc_t, I>() {
+				return cursor{__stl2::exchange(current_, __stl2::next(current_))};
```
CaseyCarter: This concerns me: can we *prove* that `current_` is valid after `next(current_)`, and that the value of `*current_` must be unchanged?

CaseyCarter: > the value of `*current_` must be unchanged

Ouch, we can't derive this for `InputIterator` now: there's no semantic requirement that `*i++` is equal to the previous value of `*i`.

ericniebler: > This concerns me: can we prove that `current_` is valid after `next(current_)`, and that the value of `*current_` must be unchanged?

My thinking is that even if `I`'s postfix increment operator is broken wrt its semantics, `move_iterator<I>`'s postfix increment operator is no _more_ broken than `I`'s.

> Ouch, we can't derive this for `InputIterator` now _with_ post increment: there's no semantic requirement that `*i++` is equal to the previous value of `*i`.

So, shall we add normative wording that _if_ the type of `i++` satisfies `Readable` that the expression `*i++` should have the same value as the expression `*i`? I can include that in my paper to nuke the Readability requirement on `i++`.

CaseyCarter: > So, shall we add normative wording that if the type of `i++` satisfies `Readable` that the expression `*i++` should have the same value as the expression `*i`?

It's only significant if we do *not* drop the `Readable` requirement on `i++`. I just want to be sure we do one or the other.

ericniebler: 👍

CaseyCarter: > can we prove that `current_` is valid after `next(current_)`, and that the value of `*current_` must be unchanged?

This cursor:

```c++
template<class T>
class cursor {
    std::istream* is_;
    std::shared_ptr<T> ptr_;
public:
    cursor() = default;
    cursor(std::istream* is) : is_{is} {}
    T const& read() const {
        return *ptr_;
    }
    void next() {
        if (!ptr_) ptr_ = std::make_shared<T>();
        if (!(*ptr_ << *is_)) is_ = nullptr;
    }
    cursor post_increment() {
        auto tmp = std::move(*this);
        next();
        return tmp;
    }
};
```
Implements something like `std::istream_iterator<T>` that uses remote state to minimize iterator copy cost for large `T`. `operator++(int)` conforms to the `Readable` requirement by saving state that is corrupted by calling `operator++()` on any copy of the cursor. I'm reasonably certain that (a) `basic_iterator<cursor<T>>` is a conforming `input_iterator`, and (b) it breaks this `move_iterator`'s `operator++(int)`.

ericniebler: Ah. So this picks the `post_increment` overload that is implemented in terms of _pre_-increment, and then we end up seeing a stomped value, is that right? OK, what if we nuked that overload of `post_increment()` and only kept the one with the proxy that is implemented in terms of _post_-fix increment?

CaseyCarter: The critical feature here is that `next(foo)` stomps on `*foo` despite operating on a copy of `foo`, which is reasonable for input iterators since they are not required to be dereferenceable after any copy is incremented.

>what if we nuked that overload of `post_increment()` and only kept the one with the proxy that is implemented in terms of post-fix increment?

I could still implement the exact same iterator without using `basic_iterator`.

ericniebler: >> what if we nuked that overload of post_increment() and only kept the one with the proxy that is implemented in terms of post-fix increment?
>
> I could still implement the exact same iterator without using `basic_iterator`.

I don't see how my suggestion is in any way effected by how `I` is implemented. The question is whether `move_iterator<I>` has the correct behavior. See the latest commit.

CaseyCarter: > I don't see how my suggestion is in any way effected by how `I` is implemented. The question is whether `move_iterator<I>` has the correct behavior. See the latest commit.

Nevermind, I completely misunderstood your comment. This version that uses `I`'s post-increment   seems almost obviously correct to me.

ericniebler: Cool, thanks for the sanity check. Squash-merge or don't, I don't care. I just wanted to make sure I understood the subtleties here.

PR #65, opened 2017-01-23
bring sanity to input and output iterators' postfix increment operators

ericniebler:
This implements the solution to the postfix increment problem that I am going to be suggesting in P0541, "Post-Increment on Input and Output Iterators". See ericniebler/stl2#232 and ericniebler/stl2#302.

For clarity, I implemented the following iterators by hand (without `basic_iterator`):
* `ostream_iterator`
* `ostreambuf_iterator`
* `insert_iterator`
* `counted_iterator`
* `common_iterator`

It's now easy to see that these are to spec except where noted (and with the exception of `noexcept` here and there).

### DIFF
--- a/include/stl2/detail/iterator/common_iterator.hpp
+++ b/include/stl2/detail/iterator/common_iterator.hpp
@@ -28,6 +28,10 @@
 // common_iterator [common.iterators]
 //
 STL2_OPEN_NAMESPACE {
+	template <Iterator I, Sentinel<I> S>
+		requires !Same<I, S>()
+	class common_iterator;
+
 	namespace __common_iterator {
 		template <class T>
 		class operator_arrow_proxy {
@@ -50,8 +54,7 @@ STL2_OPEN_NAMESPACE {
 			(std::is_pointer<I>::value ||
 				requires(const I& i) { i.operator->(); })
 		constexpr I operator_arrow_(const I& i, ext::priority_tag<2>)
-		noexcept(std::is_nothrow_copy_constructible<I>::value)
-		{
+		noexcept(std::is_nothrow_copy_constructible<I>::value) {
 			return i;
 		}
 
@@ -60,8 +63,7 @@ STL2_OPEN_NAMESPACE {
 			models::Readable<I> &&
 			_Is<reference_t<const I>, std::is_reference>
 		STL2_CONSTEXPR_EXT auto operator_arrow_(const I& i, ext::priority_tag<1>)
-		noexcept(noexcept(*i))
-		{
+		noexcept(noexcept(*i)) {
 			auto&& tmp = *i;
 			return __stl2::addressof(tmp);
 		}
@@ -76,8 +78,7 @@ STL2_OPEN_NAMESPACE {
 			std::is_nothrow_move_constructible<
 				operator_arrow_proxy<value_type_t<I>>>::value &&
 			std::is_nothrow_constructible<
-				operator_arrow_proxy<value_type_t<I>>, reference_t<I>>::value)
-		{
+				operator_arrow_proxy<value_type_t<I>>, reference_t<I>>::value) {
 			return operator_arrow_proxy<value_type_t<I>>{*i};
 		}
 
@@ -92,12 +93,8 @@ STL2_OPEN_NAMESPACE {
 			__common_iterator::operator_arrow_(i, ext::max_priority_tag)
 		)
 
-		template <Iterator I, Sentinel<I> S>
-		requires !Same<I, S>()
-		class cursor;
-
 		struct access {
-			template <_SpecializationOf<cursor> C>
+			template <_SpecializationOf<common_iterator> C>
 			static constexpr decltype(auto) v(C&& c) noexcept {
 				return (__stl2::forward<C>(c).v_);
 			}
@@ -112,6 +109,19 @@ STL2_OPEN_NAMESPACE {
 			constexpr auto operator()(const SS& s) const
 			STL2_NOEXCEPT_RETURN(
 				variant<I, S>{in_place_index<1>, s}
+			)
+		};
+
+		template <Iterator I, Sentinel<I> S, ConvertibleTo<I> II, ConvertibleTo<S> SS>
+		struct assign_visitor {
+			variant<I, S> *pv_;
+			auto operator()(const II& i) const
+			STL2_NOEXCEPT_RETURN(
+				pv_->template emplace<I>(i)
+			)
+			auto operator()(const SS& s) const
+			STL2_NOEXCEPT_RETURN(
+				pv_->template emplace<S>(s)
 			)
 		};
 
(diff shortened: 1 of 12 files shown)


include/stl2/detail/iterator/common_iterator.hpp
```diff
@@ -150,125 +160,142 @@ STL2_OPEN_NAMESPACE {
...
+				return *this;
 			}
-
-			STL2_CONSTEXPR_EXT void next()
-			noexcept(noexcept(++declval<I&>()))
-			{
+			// Not to spec:
+			// https://github.com/ericniebler/stl2/issues/232
+			decltype(auto) operator++(int)
+			noexcept(noexcept(++declval<I&>()))	{
```
ericniebler: Oops, this `noexcept` is wrong. I'll fix it tomorrow.

tahonermann:
Eric, I'm curious about the motivation for re-implementing these without the use of <tt>basic_iterator</tt>. Was this intended more to make it simpler to validate that these types match the standard, or was this based more on a belief that <tt>basic_iterator</tt> isn't up to the task for these iterators due to the post-increment requirements?  For myself, I'm still looking forward to being able to implement input/output iterators that support post-increment using <tt>basic_iterator</tt>; even if that means requiring a solution like what Casey has already implemented.

ericniebler:
> Eric, I'm curious about the motivation for re-implementing these without the use of basic_iterator. Was this intended more to make it simpler to validate that these types match the standard

This. Also, I haven't seen a clean solution yet for how to cleanly support such iterators in basic_iterator. The helpers we have are good for simple iterators, but for complicated ones, basic_iterator either doesn't do the job or else does so at the cost of obfuscation. These iterators from scratch want much more code and is far more readable IMO

include/stl2/detail/iterator/ostream_iterator.hpp
```diff
@@ -26,63 +26,58 @@
...
-				using difference_type = ostream_cursor::difference_type;
-				using char_type = charT;
-				using traits_type = traits;
-				using ostream_type = ostream_cursor::ostream_type;
-				using ebo_box<ostream_cursor>::ebo_box;
-			};
+			typedef ptrdiff_t difference_type;
+			typedef charT char_type;
+			typedef traits traits_type;
+			typedef std::basic_ostream<charT, traits> ostream_type;
```
CaseyCarter: eww, `typedef`. I will point out that these are `using` aliases in N4618 ;)

include/stl2/detail/iterator/ostream_iterator.hpp
```diff
@@ -26,63 +26,58 @@
...
+			ostream_iterator& operator++(int) {
+				return *this;
+			}
 		private:
-			raw_ptr<ostream_type> os_ = nullptr;
-			const char* delimiter_ = nullptr;
-		};
-	}
+			detail::raw_ptr<basic_ostream<charT, traits>> out_stream{nullptr};
+			const charT* delim{nullptr};
```
CaseyCarter: non-static data members have names that end with `_`. Yes, I know these are the exposition-only names given to these members in the TS. No, I don't think it will cause confusion to name them `out_stream_` and `delim_`.

include/stl2/detail/iterator/ostreambuf_iterator.hpp
```diff
@@ -23,69 +23,68 @@
...
-				using ostream_type = cursor::ostream_type;
-				using streambuf_type = cursor::streambuf_type;
-
-				constexpr mixin() noexcept = default;
-				using box_t::box_t;
+			typedef ptrdiff_t difference_type;
+			typedef charT char_type;
+			typedef traits traits_type;
+			typedef std::basic_streambuf<charT, traits> streambuf_type;
+			typedef std::basic_ostream<charT, traits> ostream_type;
```
CaseyCarter: Puke, typedefs again. If you really want these to exactly match the text in the TS, I would be happy to editorially change them to using aliases in the TS - which I think is what happened to the C++ WP wording.

include/stl2/detail/iterator/ostreambuf_iterator.hpp
```diff
@@ -23,69 +23,68 @@
...
-			: cursor{s.rdbuf()}
-			{}
-			// Extension
-			constexpr cursor(default_sentinel) noexcept
-			: cursor{}
-			{}
-
-			void write(charT c) {
+			ostreambuf_iterator() = default;
+			ostreambuf_iterator(default_sentinel) {}
```
CaseyCarter: I'd prefer:
```c++
constexpr ostreambuf_iterator(default_sentinel) noexcept
: ostreambuf_iterator{}
{}
```
for explicit future-proof paranoia.

include/stl2/detail/iterator/ostreambuf_iterator.hpp
```diff
@@ -23,69 +23,68 @@
...
-			cursor(ostream_type& s) noexcept
-			: cursor{s.rdbuf()}
-			{}
-			// Extension
-			constexpr cursor(default_sentinel) noexcept
-			: cursor{}
-			{}
-
-			void write(charT c) {
+			ostreambuf_iterator() = default;
```
CaseyCarter: I prefer `constexpr ostreambuf_iterator() noexcept = default;` so the compiler will shout loudly if we screw up this type's subobjects in the future.

include/stl2/detail/iterator/ostream_iterator.hpp
```diff
@@ -26,63 +26,58 @@
...
-				using traits_type = traits;
-				using ostream_type = ostream_cursor::ostream_type;
-				using ebo_box<ostream_cursor>::ebo_box;
-			};
+			typedef ptrdiff_t difference_type;
+			typedef charT char_type;
+			typedef traits traits_type;
+			typedef std::basic_ostream<charT, traits> ostream_type;
+
+			ostream_iterator() = default;
```
CaseyCarter: I prefer `constexpr ostream_iterator() noexcept = default;` so the compiler will shout loudly if we screw up this type's subobjects in the future.

include/stl2/detail/iterator/ostream_iterator.hpp
```diff
@@ -26,63 +26,58 @@
...
-					*os_ << delimiter_;
+			ostream_iterator& operator=(U&& u) {
+				*out_stream << static_cast<V const &>(std::forward<U>(u));
+				if (delim != nullptr) {
+					*out_stream << delim;
 				}
+				return *this;
 			}
 
+			ostream_iterator& operator*() {
```
CaseyCarter: I'd like to strengthen all three of these to `noexcept`.

include/stl2/detail/iterator/ostreambuf_iterator.hpp
```diff
@@ -23,69 +23,68 @@
...
 						sbuf_ = nullptr;
 					}
 				}
+				return *this;
 			}
-
-			// Extension
-			bool equal(const cursor& that) const noexcept {
-				return sbuf_ == that.sbuf_;
+			ostreambuf_iterator& operator*() {
```
CaseyCarter: Again here, please `noexcept` `operator*/++/++(int)`.

include/stl2/detail/iterator/ostreambuf_iterator.hpp
```diff
@@ -23,69 +23,68 @@
...
-		basic_iterator<__ostreambuf_iterator::cursor<charT, traits>>;
+			// Not to spec:
+			// https://github.com/ericniebler/stl2/issues/232
+			ostreambuf_iterator& operator++(int) {
+				return *this;
+			}
+			bool failed() const noexcept {
+				return sbuf_ != nullptr;
+			}
+			friend bool operator==(ostreambuf_iterator a, ostreambuf_iterator b) {
```
CaseyCarter: We should also `noexcept` these comparisons.

include/stl2/detail/iterator/common_iterator.hpp
```diff
@@ -150,125 +160,144 @@ STL2_OPEN_NAMESPACE {
...
 			: v_{__stl2::move(s)} {}
+
 			template <ConvertibleTo<I> II, ConvertibleTo<S> SS>
-			STL2_CONSTEXPR_EXT cursor(const cursor<II, SS>& i)
+			STL2_CONSTEXPR_EXT common_iterator(const common_iterator<II, SS>& i)
 			noexcept(noexcept(variant<I, S>{__stl2::visit(
-				convert_visitor<I, S, II, SS>{}, access::v(i))}))
-			: v_{__stl2::visit(convert_visitor<I, S, II, SS>{}, access::v(i))}
+				__common_iterator::convert_visitor<I, S, II, SS>{}, i.v_)}))
+			: v_{__stl2::visit(__common_iterator::convert_visitor<I, S, II, SS>{}, i.v_)}
```
ericniebler: Not necessary because `common_iterator<II, SS>` is a friend.

CaseyCarter: Didn't see it. I should stop reviewing code with my eyes shut.

PR #66, opened 2017-01-24
rework a bunch of the *tructible concepts, Movable requires object types

ericniebler:
Fixing a bunch of open issues with the low-level object concepts; namely, ericniebler/stl2#301, ericniebler/stl2#310, ericniebler/stl2#229, and ericniebler/stl2#70.

### DIFF
--- a/include/stl2/detail/concepts/object/move_constructible.hpp
+++ b/include/stl2/detail/concepts/object/move_constructible.hpp
@@ -26,8 +26,8 @@ STL2_OPEN_NAMESPACE {
 	constexpr bool __addressable = false;
 	template <class T>
 		requires requires(T& t, const T& ct) {
-			STL2_EXACT_TYPE_CONSTRAINT(&t, T*);
-			STL2_EXACT_TYPE_CONSTRAINT(&ct, const T*);
+			STL2_EXACT_TYPE_CONSTRAINT(&t, add_pointer_t<T>);
+			STL2_EXACT_TYPE_CONSTRAINT(&ct, add_pointer_t<const T>);
 			// Axiom: &ct == addressof(ct)
 		}
 	constexpr bool __addressable<T> = true;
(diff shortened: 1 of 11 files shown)


include/stl2/detail/concepts/object/move_constructible.hpp
```diff
@@ -25,6 +25,10 @@ STL2_OPEN_NAMESPACE {
 	template <class>
 	constexpr bool __addressable = false;
 	template <class T>
+	constexpr bool __addressable<T&> = true;
+	template <class T>
+	constexpr bool __addressable<T&&> = true;
```
CaseyCarter: This is right for `Destructible`, but wrong for `Addressable`: `Addressable` needs to be changed to `_Is<T, is_object> && __addressable<T>`. `Destructible` need not subsume `Addressable`, we can define it as:
```c++
namespace models {
  template <class T>
  constexpr bool Destructible = std::is_nothrow_destructible<T>::value && __addressable<T>;
}
template <class T>
concept bool Destructible() { return models::Destructible<T>; }
```

ericniebler: That definition of `Destructible` changes the result from `true` to `false` for reference types. We don't want that.

ericniebler: Oh, I misunderstood. You mean to leave the reference type specializations in for `__addressable`. Got it.

CaseyCarter: Yes, sorry about that.

CaseyCarter:
> Fixing a bunch of open issues 
...and ericniebler/stl2#313.

ericniebler:
...and ericniebler/stl2#300.

CaseyCarter:
...and caseycarter/stl2#22.

PR #67, opened 2017-01-25
avoid function-style casts in Common and CommonReference.

ericniebler:
This diff requires implicit convertibility to the common (reference)? type. I'n not sure that's the right thing, but it doesn't _seem_ to break anything.

### DIFF
--- a/include/stl2/type_traits.hpp
+++ b/include/stl2/type_traits.hpp
@@ -181,6 +181,8 @@ STL2_OPEN_NAMESPACE {
 		constexpr bool CommonReference<T, U> = true;
 	}
 
+	// Not to spec
+	// See https://github.com/ericniebler/stl2/issues/311
 	template <class T, class U>
 	concept bool CommonReference() {
 		return models::CommonReference<T, U>;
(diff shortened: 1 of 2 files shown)


CaseyCarter:
P0121r0 [temp.constr.conv]/1:

> An *implicit conversion constraint* is a constraint that specifies a requirement on the implicit
conversion of an expression `E` to a type `T`. The constraint is satisfied if and only if `E` is implicitly convertible to `T` (Clause 4). [ Note: A conversion constraint is introduced by a *trailing-return-type*
in a *compound-requirement* when the *trailing-return-type* contains no placeholders (5.1.4.3).
—end note ]

N4140 Clause 4 (N4618 is identical) [conv]/3:

> An expression `e` can be *implicitly converted* to a type `T` if and only if the declaration `T t=e;` is well-formed, for some invented temporary variable `t` (8.5).

`T t = e;` is not well-formed when `T` is `void` for any expression `e`.

ericniebler:
And yet `Common<void, void>` checks out.

CaseyCarter:
> And yet `Common<void, void>` checks out.

Surprise - a concepts bug.

`is_convertible<void, void>::value` is `true`; can we make it work for `Common(Reference)?`?

CaseyCarter:
See discussion in ericniebler/stl2#311.

PR #71, opened 2017-02-02
tweak Destructible, avoid inadvertant subsumption of traits with double-negation

ericniebler:


include/stl2/detail/concepts/core.hpp
```diff
@@ -20,18 +20,21 @@
 // Core Concepts [concepts.lib.corelang]
 //
 STL2_OPEN_NAMESPACE {
+	template <bool B>
+	constexpr bool __bool = B;
```
CaseyCarter: `__obfuscated_bool` ;)

PR #76, opened 2017-02-17
Massive cleanup, removing workarounds, fixing bugs, better TS conformance

ericniebler:
* Fix up Boolean, EqualityComparable, and StrictTotallyOrdered (refs ericniebler/stl2#155)
* remove argument deduction constraint work-arounds
* fix up deduction constraints (refs ericniebler/stl2#331)
* function objects implemented to spec

Tested with gcc-6.2.0 and gcc-trunk.


CaseyCarter:
Please delete the branch if you're done with it.

ericniebler:
Woohoo! :-)

PR #81, opened 2017-02-18
Small cleanups

CaseyCarter:
* Remove the `#if 0` disabled workarounds for GCC 79143.

* Implement `move_backward` and `move` algorithms directly, instead of via `copy`, `make_move_iterator`, `make_move_sentinel`, and `make_reverse_iterator`.

* ~~We can `-> Same<T>` instead of `-> Same<T>&&` when `T` is known to be `Copyable`. This is notably true for pointer types and integral types.~~

* Remove WIP `_MixinTestWrapper` from basic_iterator.hpp


CaseyCarter:
> We can `-> Same<T>` instead of `-> Same<T>&&` when `T` is known to be `Copyable`.

Not in general: `Copyable` doesn't guarantee that we can initialize from an expression of `volatile`-qualified type. That does work for fundamental types, however, including integral and pointer types - so the changes in the PR are valid.

ericniebler:
Unless I'm mistaken, a function that returns a reference to an array will satisfy -> Same<T*> because of decay.

CaseyCarter:
I do not believe you are mistaken.

ericniebler:
I'll further posit that it matters.

CaseyCarter:
> I'll further posit that it matters.

Yes, obviously.

CaseyCarter:
`UniformRandomNumberGenerator` changes reverted, there's now nothing in this patch worthy of review.

PR #83, opened 2017-02-18
fix cv-ref-qualifiers for Relation and Swappable, refs #155

ericniebler:
Also, fix a conformance problem in `Predicate`.

### DIFF
--- a/include/stl2/detail/swap.hpp
+++ b/include/stl2/detail/swap.hpp
@@ -104,27 +104,24 @@ STL2_OPEN_NAMESPACE {
 	///////////////////////////////////////////////////////////////////////////
 	// Swappable [concepts.lib.corelang.swappable]
 	//
-	template <class T, class U>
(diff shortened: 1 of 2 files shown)


include/stl2/detail/swap.hpp
```diff
@@ -104,27 +104,24 @@ STL2_OPEN_NAMESPACE {
...
-	requires
-		requires(T&& t, U&&u) {
-			__stl2::swap((T&&)t, (U&&)u);
-		}
-	constexpr bool __swappable<T, U> = true;
-
 	template <class T>
 	concept bool Swappable() {
-		return __swappable<T, T>;
+		return requires(T&& t, T&&u) {
```
CaseyCarter: `T&&u`! It burns! Could we make these `x` and `y`, or `a` and `b`?

ericniebler: Picky picky. ;-)

Issue #69, opened 2017-01-30
Eradicate initializer_list algorithm extensions


CaseyCarter:
...being careful not to remove the non-extension overloads that accept `initializer_list`.

cjdb:
Are there any extensions that don't have `// .*Extension.*` above them, and are there any non-extensions that _might_ have `// .*Extension.*` above them?

cjdb:
> Are there any extensions that don't have `// .*Extension.*` above them

Yes, evidenced by `count_if.hpp`.

CaseyCarter:
> Yes, evidenced by count_if.hpp.

Curses, I thought I had managed to tag all of them. In any case, `min`, `max`, and `minmax` are the only algorithms specified to have `initializer_list` overloads.

cjdb:
Hmm... I just checked out all of my edits, because they fail tests, and I thought I'd royally screwed up. Do I need to remove the tests too in that case?

cjdb:
Previously on #69:

> Are there any algorithms that don't have `// .*Extension.*` that are in fact extensions?
> Yes, evidenced by `count_if.hpp`

CaseyCarter:
> Do I need to remove the tests too in that case?

I'd be wary of inadvertently reducing test coverage. Maybe change the tests to use lvalue initializer lists a la:

```diff
-algorithm(foo, {1,2,3});
+auto some_list = {1,2,3};
+algorithm(foo, some_list);
```

cjdb:
Got it, thanks for the heads up :)

PR #84, opened 2017-02-18
Patch for #69

cjdb:
Tested on gcc-6 (Bash for Windows) and gcc trunk (Linux).


test/algorithm/count_if.cpp
```diff
@@ -78,8 +78,14 @@ int main()
 	CHECK(count_if(make_range(input_iterator<T*>(ta),
 						 sentinel<T*>(ta + size(ta))), &T::b) == 4);
 
-	CHECK(count_if({0, 1, 2, 2, 0, 1, 2, 3}, equals(2)) == 3);
-	CHECK(count_if({0, 1, 2, 2, 0, 1, 2, 3}, equals(42)) == 0);
+	{
+		auto l = {0, 1, 2, 2, 0, 1, 2, 3};
+		CHECK(count_if(std::move(l), equals(2)) == 3);
+   }
+   {
```
CaseyCarter: 4-space indentation on these two lines should be tab.

cjdb: Ahh, whoops! I really should configure Notepad++ to do tabs :)

test/algorithm/count_if.cpp
```diff
@@ -81,8 +81,8 @@ int main()
 	{
 		auto l = {0, 1, 2, 2, 0, 1, 2, 3};
 		CHECK(count_if(std::move(l), equals(2)) == 3);
-   }
-   {
+		}
+		{
```
CaseyCarter: I think you meant s/    /\t/g ;)

PR #90, opened 2017-02-28
Adjust Readable and Writable per P0547.

CaseyCarter:


include/stl2/detail/iterator/concepts.hpp
```diff
@@ -206,8 +205,7 @@ STL2_OPEN_NAMESPACE {
 
 	template <class Out, class R>
 	concept bool Writable() {
-		return detail::Dereferenceable<Out> && Movable<Out>() &&
-			DefaultConstructible<Out>() && __writable<Out, R>;
+		return detail::Dereferenceable<Out> && __writable<Out, R>;
```
ericniebler: Suggest killing `__writable`.

PR #91, opened 2017-03-12
Make iterators that return by value non-Writable

ericniebler:
refs ericniebler/stl2#381, refs ericniebler/range-v3#573

Interestingly, none of the standard output iterators (ostream(buf) or the various insert iterators) are affected since their `operator*` returns `*this` by reference. :-)

This fix works out quite well in practice. More generally, I feel like testing for the writability of a `const`-qualified prvalue could play a part in the resolution of ericniebler/stl2#226. We could make it the hallmark of proxy reference types. (That still doesn't help us with specifying the *semantics* of the various operations on proxy references, but at least we have an inkling of how we might specify the syntax.)

### DIFF
--- a/include/stl2/detail/iterator/concepts.hpp
+++ b/include/stl2/detail/iterator/concepts.hpp
@@ -186,8 +186,8 @@ STL2_OPEN_NAMESPACE {
 	///////////////////////////////////////////////////////////////////////////
 	// Writable [iterators.writable]
 	// Not to spec, but getting closer.
-	// See https://github.com/ericniebler/stl2/issues/240
-	// Additional changes from P0547
+	// See https://github.com/ericniebler/stl2/issues/339
+	// and https://github.com/ericniebler/stl2/issues/381
 	//
 	template <class Out, class R>
 	concept bool Writable() {
@@ -198,6 +198,8 @@ STL2_OPEN_NAMESPACE {
 			// Axiom: If r equals foo && Readable<Out>() &&
 			//        Same<value_type_t<Out>, ????>() then
 			//        (*o = (R&&)r, *o equals foo)
+			(void)(const_cast<const reference_t<Out>&&>(*o) = (R&&)r);
+			// Axiom: This expression is equivalent to the previous expression.
 		};
 	}
 
--- a/test/iterator/basic_iterator.cpp
+++ b/test/iterator/basic_iterator.cpp
@@ -239,6 +239,20 @@ struct proxy_wrapper {
 		return *this;
 	}
 
+	proxy_wrapper const& operator=(const T& t) const
+	noexcept(std::is_nothrow_copy_assignable<T>::value)
+	{
+		get() = t;
+		return *this;
+	}
+
+	proxy_wrapper const& operator=(T&& t) const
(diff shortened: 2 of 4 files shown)


ericniebler:
FYI, this is the direction I've taken in range-v3: ericniebler/range-v3@f522f421c055bf3f31fca19b7325b549a6a5f318.

include/stl2/detail/iterator/concepts.hpp
```diff
@@ -186,13 +186,14 @@ STL2_OPEN_NAMESPACE {
...
-	// See https://github.com/ericniebler/stl2/issues/240
-	// Additional changes from P0547
+	// See https://github.com/ericniebler/stl2/issues/339
+	// and https://github.com/ericniebler/stl2/issues/381
 	//
 	template <class Out, class R>
 	concept bool Writable() {
 		return detail::Dereferenceable<Out> && requires(Out& o, R&& r) {
 			(void)(*o = (R&&)r);
+			(void)(const_cast<add_const_t<reference_t<Out>>&&>(*o) = (R&&)r);
```
CaseyCarter: I'd prefer `const` here to `add_const_t`, IIRC they're equivalent for referenceable types and `Dereferenceable` guarantees that `reference_t<Out>` is a referenceable type.

test/iterator/basic_iterator.cpp
```diff
@@ -239,6 +239,20 @@ struct proxy_wrapper {
...
+		return *this;
+	}
+
+	proxy_wrapper const& operator=(T&& t) const
+	noexcept(std::is_nothrow_move_assignable<T>::value)
+	{
+		get() = stl2::move(t);
+		return *this;
+	}
+
```
CaseyCarter: Do we need the non-`const` overloads anymore?

test/iterator/iterator.cpp
```diff
@@ -33,13 +33,23 @@ struct reference_wrapper {
...
+		return *this;
+	}
 
-	reference_wrapper& operator=(T&& t)
+	reference_wrapper const& operator=(const T& t) const
+		noexcept(std::is_nothrow_copy_assignable<T>::value) {
+		get() = t;
+		return *this;
+	}
+	reference_wrapper const& operator=(T&& t) const
```
CaseyCarter: Same comment: we don't need to satisfy `Assignable`; I think the non-`const` overloads are extraneous.

test/iterator/basic_iterator.cpp
ericniebler: I got strange ambiguities when I removed the non-const overloads, but didn't bother to track it down. Maybe it has something to do with the presence of the implicit conversion operator? I dunno. I was lazy.

ericniebler:
I realize that I also need to change `basic_proxy_reference` to add `const` overloads of `operator=`. Test coverage of `basic_iterator` is light now that I've switched the iterator adaptors to be specified by hand.

test/iterator/basic_iterator.cpp
CaseyCarter: File an issue for someone to look into it if/when we feel the need; it's not important enough to me to block the PR.

PR #96, opened 2017-04-21
Use Travis CI's "trusty" containerized build environment

CaseyCarter:


ericniebler:
Why do the travis results for this only show gcc-6 debug and release?

CaseyCarter:
Because this repo only runs gcc-6 debug and release. Are you lost?

ericniebler:
> Are you lost?

Yep.

PR #98, opened 2017-04-27
Fixes #97: Changed include guard from "REMOVE_HPP" to "REMOVE_IF_HPP"

### LINKED ISSUE #97: remove_if.hpp include guard is same as remove.hpp

cjdb:


CaseyCarter:
Nice catch.

PR #100, opened 2017-05-30
fix concept checks for const counted_ and common_iterator

ericniebler:
refs ericniebler/stl2#368

### DIFF
--- a/include/stl2/detail/iterator/common_iterator.hpp
+++ b/include/stl2/detail/iterator/common_iterator.hpp
@@ -211,13 +211,14 @@ STL2_OPEN_NAMESPACE {
 			return *__stl2::get_unchecked<I>(v_);
 		}
 		decltype(auto) operator*() const
-		noexcept(noexcept(*std::declval<const I&>())) {
+		noexcept(noexcept(*std::declval<const I&>()))
+		requires detail::Dereferenceable<const I> {
 			STL2_EXPECT(__stl2::holds_alternative<I>(v_));
 			return *__stl2::get_unchecked<I>(v_);
 		}
 		decltype(auto) operator->() const
 		noexcept(noexcept(__common_iterator::operator_arrow(std::declval<const I&>())))
-		requires Readable<I>() {
+		requires Readable<const I>() {
 			STL2_EXPECT(__stl2::holds_alternative<I>(v_));
 			return __common_iterator::operator_arrow(__stl2::get_unchecked<I>(v_));
 		}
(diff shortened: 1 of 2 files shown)


include/stl2/detail/iterator/common_iterator.hpp
```diff
@@ -211,13 +211,14 @@ STL2_OPEN_NAMESPACE {
 			return *__stl2::get_unchecked<I>(v_);
 		}
 		decltype(auto) operator*() const
-		noexcept(noexcept(*std::declval<const I&>())) {
+		noexcept(noexcept(*std::declval<const I&>()))
+		requires detail::Dereferenceable<const I&> {
```
CaseyCarter: `Dereferenceable<const I>` (the concept is oblivious to value category)

include/stl2/detail/iterator/common_iterator.hpp
```diff
@@ -211,13 +211,14 @@ STL2_OPEN_NAMESPACE {
...
 			STL2_EXPECT(__stl2::holds_alternative<I>(v_));
 			return *__stl2::get_unchecked<I>(v_);
 		}
 		decltype(auto) operator->() const
 		noexcept(noexcept(__common_iterator::operator_arrow(std::declval<const I&>())))
-		requires Readable<I>() {
+		requires Readable<const I>() {
```
CaseyCarter: `Readable<I>` indirectly requires `Dereferenceable<const I>`, which should suffice here. `Readable<const I>` requires `Movable<const I>` requires `Assignable<const I&, const I>` which is certainly not right.

include/stl2/detail/iterator/counted_iterator.hpp
```diff
@@ -66,7 +66,8 @@ STL2_OPEN_NAMESPACE {
 			return *current();
 		}
 		decltype(auto) operator*() const
-			noexcept(noexcept(*std::declval<const I&>())) {
+		noexcept(noexcept(*std::declval<const I&>()))
+		requires detail::Dereferenceable<const I&> {
```
CaseyCarter: Ditto, `detail::Dereferenceable<const I>`.

include/stl2/detail/iterator/common_iterator.hpp
ericniebler: Not after the adoption of ericniebler/stl2#339.

CaseyCarter: Aha! There it is - I was certain we had futzed with `Readable` somewhere but could not find it.

ericniebler: I'm a bit concerned that we have too many issue resolutions in flight, and something like this will get lost in the shuffle. Although I suppose that won't happen if we're diligent about out issue tagging.

Maybe we should have open issues for the papers, too, so we don't forget to merge them.

PR #101, opened 2017-05-31
switch to variable-like concepts

ericniebler:
refs ericniebler/stl2#379


ericniebler:
A few hours of work, but I believe it to be worth it.

ericniebler:
I take it you approve of the direction. Now I just have to finish the paper. :-P

PR #109, opened 2017-06-26
stl2 common_type dispatches to stl1 common_type

ericniebler:
suggested fix for ericniebler/stl2#421


CaseyCarter:
We also need to change the specializations of `s::e::r::common_type` to specializations of `std::common_type`.

PR #116, opened 2017-07-31
Implement span

CaseyCarter:


CaseyCarter:
This needs a quick review to be sure it's consistent with range-v3 span.

PR #121, opened 2017-10-18
Range adaptors ts misc

CaseyCarter:
Slew of mostly small cleanup / style changes.

### DIFF
--- a/include/stl2/detail/range/concepts.hpp
+++ b/include/stl2/detail/range/concepts.hpp
@@ -108,15 +108,14 @@ STL2_OPEN_NAMESPACE {
 	template <class T>
 	struct __view_predicate2 : __view_predicate3<T> {};
 
-	template <class T>
-		requires meta::_v<std::is_base_of<view_base, T>>
+	template <DerivedFrom<view_base> T>
 	struct __view_predicate2<T> : std::true_type {};
 
 	template <class T>
 	struct __view_predicate : __view_predicate2<T> {};
 
 	template <class T>
-		requires _Valid<meta::_t, enable_view<T>>
+		requires meta::Trait<enable_view<T>>
 	struct __view_predicate<T> : enable_view<T> {};
 
 	template <class T>
--- a/include/stl2/detail/semiregular_box.hpp
+++ b/include/stl2/detail/semiregular_box.hpp
@@ -12,17 +12,19 @@
 #ifndef STL2_DETAIL_SEMIREGULAR_BOX_HPP
 #define STL2_DETAIL_SEMIREGULAR_BOX_HPP
 
+#include <type_traits>
 #include <stl2/optional.hpp>
-#include <stl2/detail/cheap_storage.hpp>
+#include <stl2/detail/ebo_box.hpp>
 #include <stl2/detail/fwd.hpp>
-#include <stl2/detail/meta.hpp>
 #include <stl2/detail/concepts/object.hpp>
 
 STL2_OPEN_NAMESPACE {
 	namespace detail {
-		template <CopyConstructible T>
-		class semiregular_box {
-		public:
+		// If T models MoveConstructible, semiregular_box<T> models Movable &&
+		// DefaultConstructible (so-called "MoveSemiregular"). If T models
+		// CopyConstructible, semiregular_box<T> models Semiregular.
+		template <ext::MoveConstructibleObject T>
+		struct semiregular_box {
 			semiregular_box() = default;
 			constexpr semiregular_box()
 			noexcept(is_nothrow_default_constructible<T>::value)
@@ -35,7 +37,11 @@ STL2_OPEN_NAMESPACE {
 
 			constexpr semiregular_box(const T& t)
 			noexcept(is_nothrow_copy_constructible<T>::value)
+			requires CopyConstructible<T>
 			: o_{in_place, t} {}
+
+			semiregular_box(const semiregular_box&) requires CopyConstructible<T> = default;
+			semiregular_box(semiregular_box&&) = default;
 
 			template <class... Args>
 			requires Constructible<T, Args...>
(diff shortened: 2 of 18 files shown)


include/stl2/detail/range/concepts.hpp
```diff
@@ -108,15 +108,14 @@ STL2_OPEN_NAMESPACE {
 	template <class T>
 	struct __view_predicate2 : __view_predicate3<T> {};
 
-	template <class T>
-		requires meta::_v<std::is_base_of<view_base, T>>
+	template <DerivedFrom<view_base> T>
```
ericniebler: Doesn't this require "publicly and unambiguously"? I don't think we want that.

include/stl2/detail/semiregular_box.hpp
```diff
@@ -35,61 +37,38 @@ STL2_OPEN_NAMESPACE {
...
+
 			constexpr semiregular_box& operator=(semiregular_box&& that) &
 			noexcept(is_nothrow_move_constructible<T>::value)
 			{
+				o_.reset();
 				if (that.o_)
 					o_.emplace(std::move(*that.o_));
-				else
-					o_.reset();
-				return *this;
```
ericniebler: Why force a destruct/reconstruct when an assignment might be more efficient?

include/stl2/detail/view/view_closure.hpp
```diff
@@ -23,7 +22,7 @@ STL2_OPEN_NAMESPACE {
 
 		template <class T>
 		concept bool Pipeable =
-			std::is_base_of_v<__pipeable_base, T> &&
+			DerivedFrom<T, __pipeable_base> &&
```
ericniebler: Again, I don't think we want "publicly and unambiguously".

include/stl2/view/all.hpp
```diff
@@ -53,9 +52,8 @@ STL2_OPEN_NAMESPACE {
 	}
 
 	namespace ext {
-		template <Range Rng>
-		requires std::is_lvalue_reference_v<Rng> || View<decay_t<Rng>>
-		using all_view = std::decay_t<decltype(view::all(std::declval<Rng>()))>;
+		template <class Rng>
+		using all_view = decltype(view::all(std::declval<Rng>()));
```
ericniebler: I prefer to leave the explicit `requires` clause. It will give better error messages.

include/stl2/view/iota.hpp
```diff
@@ -68,23 +68,16 @@ STL2_OPEN_NAMESPACE {
...
-		explicit iota_view(I value) -> iota_view<I>;
-
-		template <Incrementable I>
-		iota_view(I value, I bound) -> iota_view<I, I>;
-
-		template <WeaklyIncrementable I, Semiregular Bound>
-		requires WeaklyEqualityComparable<I, Bound> && !ConvertibleTo<Bound, I>
-		iota_view(I value, Bound bound) -> iota_view<I, Bound>;
+		template <Incrementable I, ConvertibleTo<I> Bound>
+		iota_view(I value, Bound) -> iota_view<I, I>;
```
ericniebler: No, this is _not_ semantically equivalent, and it's not what I want. I want `iota_view{0, v.size()}` to be an error. Users should have to disambiguate with a cast: `iota_view{0, (int)v.size()}`, or some such.

include/stl2/view/view_interface.hpp
```diff
@@ -43,12 +43,12 @@ STL2_OPEN_NAMESPACE {
 			Constructible<Cont, __range_common_iterator<Rng>, __range_common_iterator<Rng>>;
 
 		template <class D>
-		class view_interface : __stl2::view_base {
+		class view_interface : public view_base {
```
ericniebler: This shouldn't have to be public.

include/stl2/view/view_interface.hpp
```diff
@@ -94,14 +94,14 @@ STL2_OPEN_NAMESPACE {
 			constexpr decltype(auto) at(difference_type_t<iterator_t<R>> n) {
 				return n < 0 || n >= __stl2::size(derived())
 					? throw std::out_of_range{}
-					: (*this)[n];
+					: derived()[n];
```
ericniebler: Why?

include/stl2/detail/range/concepts.hpp
CaseyCarter: Yes it does, and intentionally: `std::tuple<int, double, iota_view<int, int>>` is NOT a view.

include/stl2/detail/semiregular_box.hpp
CaseyCarter: This is only used for types that are not move assignable; they get the `requires Movable<T>` defaulted overload below

include/stl2/detail/view/view_closure.hpp
CaseyCarter: Again, I do ;)

include/stl2/view/view_interface.hpp
CaseyCarter: The derived type might implement `operator[]`.

include/stl2/view/all.hpp
CaseyCarter: Yeah, I overreacted here when I noticed that `View<decay_t<Rng>>` should be `View<__f<Rng>>`. Will fix.

include/stl2/view/iota.hpp
CaseyCarter: Yes, we discussed that in the other PR. I'll remove this guide.

CaseyCarter: > I'll remove this guide.

...and kill the implicit guide for the `iota_view(I, Bound)` constructor, and restore the original set of explicit guides, so that you get your ambiguity.

include/stl2/view/iota.hpp
```diff
@@ -52,8 +52,7 @@ STL2_OPEN_NAMESPACE {
 			constexpr explicit iota_view(I value) requires Same<Bound, unreachable>
 			: value_(value), bound_{} {}
 			template <class II = I, class BB = Bound>
-			requires Same<I, II> && Same<Bound, BB>
-			constexpr iota_view(II value, BB bound)
+			constexpr iota_view(meta::id_t<II> value, meta::id_t<BB> bound)
```
ericniebler: What the hell is this?

CaseyCarter: A constructor with effectively no implicit deduction guide, so that `iota_view{0, std::size_t{1}}` fails class template argument deduction.

ericniebler: *_*HATE.*_*

tahonermann: Sold.  I now hate implicit deduction guides too.

CaseyCarter: > Sold. I now hate implicit deduction guides too.

We *really* need the ability to `delete` guides.

include/stl2/view/join.hpp
```diff
@@ -1,6 +1,6 @@
 // cmcstl2 - A concept-enabled C++ standard library
 //
-//  Copyright Casey Carter 2016
+//  Copyright Eric Niebler 2017
```
ericniebler: uups, thanks

PR #126, opened 2017-11-06
implement subrange per Casey's review changes

CaseyCarter:


ericniebler:
Do any of the `subrange` changes need to be posted to P0789?

include/stl2/view/all.hpp
```diff
@@ -46,7 +47,7 @@ STL2_OPEN_NAMESPACE {
 	// Work-around for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82507
 	template <class V, class R>
 	concept bool _ConstructibleFromRange =
-		View<V> && ext::ViewableRange<R> && Constructible<V, ext::all_view<R>>;
+		View<V> && Constructible<V, ext::all_view<R>>;
```
CaseyCarter: This concept definition violates the "first instantiate types in a `typename` constraint" policy.

CaseyCarter:
> Do any of the `subrange` changes need to be ported to P0789?

Yes, absolutely - I was wanting to get your feedback on `ReferenceableRange`/`is_referenceable_range`/`safe_subrange_t` (described in the commit message for caseycarter/range-v3@8a2102f5cbe4d764b5f2098beafa7ed245a36f7d) before I devise wording for `ref_view`, the modified `subrange` converting constructors/deduction guides and necessary changes to `view::all`.

include/stl2/view/subrange.hpp
```diff
@@ -12,260 +12,245 @@
...
-
-			template <ConvertibleTo<I> X, ConvertibleTo<S> Y, bool B>
-			constexpr subrange(subrange<X, Y, B> r, difference_type_t<I> n)
-			requires __bool<Sized> // strange GCC bug, not yet reported
-			: subrange{r.begin(), r.end(), n} {}
-
-			template <_PairLikeConvertibleTo<I, S> PairLike>
-			constexpr subrange(PairLike r)
-			: subrange{std::get<0>(std::move(r)), std::get<1>(std::move(r))} {}
+		enum subrange_kind : bool { unsized, sized };
```
CaseyCarter: `enum class`

ericniebler: Maybe we want to make this `: byte` instead of `: bool` so that we can extend this in the future without breaking ABI.

include/stl2/view/subrange.hpp
```diff
@@ -12,260 +12,245 @@
...
-		meta::Integral<std::tuple_size<__uncvref<T>>> &&
-		std::tuple_size<__uncvref<T>>::value == 2 &&
-		meta::Trait<std::tuple_element<0, __uncvref<T>>> &&
-		meta::Trait<std::tuple_element<1, __uncvref<T>>>;
+	concept bool _PairLike = requires(T t) {
+		typename std::tuple_size<T>;
+		requires meta::Integral<std::tuple_size<T>>;
+		requires std::tuple_size<T>::value == 2;
+		typename std::tuple_element<0, T>;
+		typename std::tuple_element<1, T>;
```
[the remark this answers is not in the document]
CaseyCarter: No; and the answer's the same for `tuple_size` above. They may be incomplete, but they're always types since there are no constraints on the base template. This is too slavishly following the "always form types first in a typename constraint" rule.

ericniebler:
> I'm uncomfortable with `ReferenceableRange` and the special-casing of `subrange` and `ref_view`. It feels too ad hoc to me.

It would be nice if we could get to a place where <tt>begin(<em>range-expr</em>)</tt> returned <tt>dangling&lt;<em>iterator-type</em>&gt;</tt> if the result were potentially dangling. That is, we remove the deprecated `Rng const&` overloads of `begin()` and `end()` and replace them with generic `Rng const &&` overloads that wrap the return values in `dangling<>`. Then we can add overloads of `begin()` and `end()` for rvalue `subrange<>` and `ref_view<>` that return non-dangling iterators. The difference would be testable at compile-time. This seems to me to be more in the spirit of the design than an extra `is_referenceable_range` customization point.

It would be surprising if `Range<std::vector<int>>` wasn't satisfied so we would need to think through how the concept plays with such a scheme. It currently would work fine since `iterator_t` and `sentinel_t` test with lvalues, but we would need to make sure everything is still kosher.

Does this make sense? Thoughts?

CaseyCarter:
> I'm uncomfortable with `ReferenceableRange` and the special-casing of `subrange` and `ref_view`. It feels too ad hoc to me. The semantics of `Range` and `View` -- which are subtle to begin with -- are getting muddied by this thing in the middle. It sets my Generic Programming spidey sense tingling.

While it certainly *looks* ad-hoc - there's a concept here with effectively no requirements but an opt-in trait - I assure you this is actually principled. `ReferenceableRange` is a concept that's been around since 2014: it is a refinement of `Range` that characterizes ranges which can be passed to algorithms whose returned iterators are guaranteed not to dangle. Previously, `ReferenceableRange` - as embodied by the `safe_iterator_t` alias - only admitted lvalue reference types. This change recognizes that there are rvalue range types which satisfy the requirements of `ReferenceableRange`.

The motivation behind admitting lvalues is that (non-`const`) lvalue references will not bind to temporaries. When an algorithm accepts an lvalue reference to a range there is a reasonable expectation that the range will not be destroyed at the end of the full expression so that returned iterators remain valid. The existence of `as_lvalue` in range-v3 attests that this property is heuristic: it defends against Murphy instead of Machiavelli. A determined caller can provide an lvalue reference to a range that will be destroyed at the end of the full expression leaving returned iterators dangling, but it's not possible to do so inadvertently.

The salient feature of lvalue references that prevents their iterators dangling is that the validity of iterators (and sentinels) obtained from an lvalue-reference-to-range is not tied to the lifetime of the reference, but to the lifetime of the referenced model of range. The iterators of the lvalue-reference-to-range are exactly the iterators of the referenced range. Put simply, they have reference semantics. 

It's fairly straight-forward to see that `ref_view` has similar properties. Like an lvalue-reference-to-range, a `ref_view` can only be constructed from a `ReferenceableRange`. `ref_view` also uses the iterators (and sentinel) of the referenced range as its own iterators. Again, the validity of the obtained iterators is in no way tied to the lifetime of the `ref_view` - validity depends only on the lifetime of the underlying object referenced. (None of this should be surprising: `ref_view` is by design a `Semiregular` type that is roughly isomorphic to lvalue-reference-to-range.) Given that it has the same properties, `ref_view` provides as much guarantee as "lvalue-reference-to-range" that returned iterators do not dangle. It is sensible to admit `ref_view` into `ReferenceableRange`.

The story for `subrange` is similar albeit with a few twists. A `subrange` can be constructed either from a `ReferenceableRange` or from an iterator and sentinel value. Since `begin` and `end` (modulo the deprecated behavior for rvalues) require lvalue arguments, the iterator and sentinel used to construct a `subrange` either came from a `ReferenceableRange`, or have validity that does not depend on the lifetime of any external range object (as is common for pre-Ranges-TS iterators). In either case, the validity of iterators obtained from the `subrange` does not depend on the lifetime of the `subrange` - it depends on the lifetime of the `ReferenceableRange` from which the iterator used to construct the `subrange` was originally obtained. (I'm deliberately ignoring the pair converting constructor here, which I believe needs to be made `explicit`. Even absent anything to do with `ReferenceableRange`, I think converting pairs of iterators and/or sentinels to ranges should require explicit action from a user, for the same reasons that we didn't add `begin` and `end` overloads for `pair`.)

With all that said, I agree that the name `ReferenceableRange` is too vague and that its expression as a concept here is janky. It occurred to me hours after pushing this that the concept lacks syntactic distinction from `Range` and that the semantic properties will be hard to express. Making the customization point user-specializable without any kind of syntax guards is a recipe for everyone specializing the trait and trashing the lifetime semantics of the design. `is_referenceable_range` should be an implementation detail - at least until the concept is better developed.

CaseyCarter:
> It would be nice if we could get to a place where begin(range-expr) returned dangling<iterator-type> if the result were potentially dangling. That is, we remove the deprecated Rng const& overloads of begin() and end() and replace them with generic Rng const && overloads that wrap the return values in dangling<>. Then we can add overloads of begin() and end() for rvalue subrange<> and ref_view<> that return non-dangling iterators. The difference would be testable at compile-time. This seems to me to be more in the spirit of the design than an extra is_referenceable_range customization point.

I believe it is sensible for `begin` and `end` to accept `ReferenceableRange`, although I haven't given any thought to consequences. I'm less enthusiastic about `begin` and `end` returning `dangling` types for non-`ReferenceableRange` arguments: I would not want to lose the guarantee that `begin` returns an iterator and `end` a sentinel for that iterator without seeing a clear benefit. 

>
>It would be surprising if `Range<std::vector<int>>` wasn't satisfied so we would need to think through how the concept plays with such a scheme. It currently would work fine since `iterator_t` and `sentinel_t` test with lvalues, but we would need to make sure everything is still kosher.

I think this scheme has the potential to turn nearly all `Range`s (or `View`s, if we try to restrict the effect to non-containers) into `ReferenceableRange`s: `&`-qualified `begin` and `end` are incredibly rare in the wild, and I'm not sure how we'd distinguish between the `ReferenceableRange` `subrange::begin` accepting rvalues and the non-`ReferenceableRange` `std::basic_string_view::begin` accepting rvalues.

(Given my statement above about needing to develop syntactic requirements for `ReferenceableRange` to allow it to be user-extendable I'm not totally opposed to this idea, I just not certain this is the right approach.)

ericniebler:
> I think this scheme has the potential to turn nearly all Ranges (or Views, if we try to restrict the effect to non-containers) into `ReferenceableRanges`: `&`-qualified begin and end are incredibly rare in the wild

That also occurred to me. The solution I was toying with is to change the `begin` and `end` CPOs to only look for `begin`/`end` _free_ functions via ADL when the argument is an rvalue. The onus is then on authors of `subrange`-like types to write such functions as an opt-in. It's a bit of a hack, but better (I think) than adding a new customization point.

The risk then becomes types for which users have already written `begin`/`end` ADL customization points with `&`/`const &` overloads, as that will make it appear to be "rvalue-safe". But for many (most?) of those types, the types of the begin and end iterators will depend on the const-ness of the arguments, making them containers and therefor not rvalue-safe.

CaseyCarter:
> That also occurred to me. The solution I was toying with is to change the begin and end CPOs to only look for begin/end free functions via ADL when the argument is an rvalue. The onus is then on authors of subrange-like types to write such functions as an opt-in. It's a bit of a hack, but better (I think) than adding a new customization point.

This is a dark enough corner that the weird hack may be justifiable: we really don't want to make it too easy to use, or it will be easy to mis-use as well. I could live with this as an extension mechanism.

ericniebler:
> > That also occurred to me. The solution I was toying with is to change the begin and end CPOs to only look for begin/end free functions via ADL when the argument is an rvalue. The onus is then on authors of subrange-like types to write such functions as an opt-in. It's a bit of a hack, but better (I think) than adding a new customization point.
>
> This is a dark enough corner that the weird hack may be justifiable: we really don't want to make it too easy to use, or it will be easy to mis-use as well. I could live with this as an extension mechanism.

Actually, I think we get what we want by changing the poison pill as follows:

```diff
-void begin(auto&) = delete;
+void begin(auto&&) = delete;
```

Then, after changing the CPO to only consider ADL `begin` functions for rvalues, we avoid the problem of accidental opt-in by `const&` overloads of `begin`.

See https://godbolt.org/g/3xc8rp

PR #128, opened 2017-11-10
Support CMake find_package(cmcstl2)

johelegp:
This PR allows consuming the library in an idiomatic CMake way. Previously, I had to hardcode "include(/usr/local/cmcstl2/cmcstl2.cmake)". Now we can do "find_package(cmcstl2)".

Greetings,
Johel


CaseyCarter:
Thanks for the contribution, Johel - this looks great.

PR #134, opened 2018-02-16
redesigned range access CPOs remove unsafe deprecated behavior

ericniebler:
also, make it possible for a range type to opt-in to safe rvalue range-ness

### DIFF
--- a/include/stl2/detail/range/access.hpp
+++ b/include/stl2/detail/range/access.hpp
@@ -29,22 +29,25 @@ STL2_OPEN_NAMESPACE {
 	namespace __begin {
 		// Poison pill for std::begin. (See the detailed discussion at
 		// https://github.com/ericniebler/stl2/issues/139)
-		void begin(auto&) = delete;
+		// Not to spec
+		template <class R>
+		void begin(R&&) = delete;
 
 		template <class R>
 		constexpr bool has_member = false;
 		template <class R>
 			requires requires(R& r) {
+				//{ r.begin() } -> Iterator;
 				requires Iterator<__f<decltype(r.begin())>>;
 			}
-		constexpr bool has_member<R> = true;
+		constexpr bool has_member<R&> = true;
 
 		template <class R>
 		constexpr bool has_non_member = false;
 		template <class R>
-		requires
-			requires(R& r) {
-				requires Iterator<__f<decltype(begin(r))>>;
+		requires requires(R&& r) {
+				// { begin((R&&) r) } -> Iterator;
+				requires Iterator<__f<decltype(begin((R&&) r))>>;
 			}
 		constexpr bool has_non_member<R> = true;
 
@@ -55,28 +58,18 @@ STL2_OPEN_NAMESPACE {
 			}
 			// Prefer member if it returns Iterator.
 			template <class R>
-			requires has_member<R>
-			constexpr Iterator
-			operator()(R& r) const
+			requires has_member<R&>
+			constexpr auto operator()(R& r) const
 			STL2_NOEXCEPT_RETURN(
 				r.begin()
 			)
 			// Use ADL if it returns Iterator.
 			template <class R>
 			requires !has_member<R> && has_non_member<R>
-			constexpr Iterator
-			operator()(R& r) const
+			constexpr auto operator()(R&& r) const
 			STL2_NOEXCEPT_RETURN(
-				begin(r)
+				begin((R&&) r)
 			)
-			template <_IsNot<is_array> R>
-			[[deprecated]] constexpr Iterator
-			operator()(const R&& r) const
-			noexcept(noexcept(declval<const fn&>()(r)))
-			requires has_member<const R> || has_non_member<const R>
-			{
-				return (*this)(r);
-			}
 		};
 	}
 	// Workaround GCC PR66957 by declaring this unnamed namespace inline.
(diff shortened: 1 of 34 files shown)


ericniebler:
Tested locally with gcc trunk.

ericniebler:
The next step is to update `safe_iterator_t` to take advantage of the extra information afforded by the new range access customization points.

include/stl2/detail/range/access.hpp
```diff
@@ -415,49 +449,42 @@ STL2_OPEN_NAMESPACE {
 		template <class R>
 		constexpr bool has_member = false;
 		template <class R>
-		requires
-			requires(const R& r) {
-				{ r.empty() } -> bool;
+		requires requires(const R& r) {
+				// { r.empty() } -> bool;
+				requires ConvertibleTo<__f<decltype(r.empty())>, bool>;
```
CaseyCarter: This is not to spec (and neither was `{ r.empty() } -> bool;`). This should be a simple expression requirement `bool(r.empty())`.

include/stl2/detail/range/access.hpp
```diff
@@ -415,49 +449,42 @@ STL2_OPEN_NAMESPACE {
...
 
 		struct fn {
 			// Prefer member
 			template <class R>
-			requires
-				has_member<R>
+			requires has_member<R>
 			constexpr bool operator()(const R& r) const
 			STL2_NOEXCEPT_RETURN(
 				static_cast<bool>(r.empty())
```
CaseyCarter: Admittedly, this is not exactly `bool(r.empty())`, but I *think* the `static_cast` is equivalent here (perhaps only because `bool` is not a class type).

include/stl2/detail/range/access.hpp
```diff
@@ -474,76 +501,58 @@ STL2_OPEN_NAMESPACE {
...
 			//            iterators.
 			template <class R>
 			requires !has_member<R> && !has_pointer_iterator<R> &&
 				has_contiguous_iterator<R>
-			constexpr auto operator()(R& r) const
-			noexcept(noexcept(__stl2::begin(r) == __stl2::end(r)
-				? nullptr : detail::addressof(*__stl2::begin(r))))
+			constexpr auto operator()(R&& r) const
+			noexcept(noexcept(*declval<__begin_t<R>&>()) &&
+				noexcept(declval<__begin_t<R>&>() == declval<__end_t<R>>()))
```
CaseyCarter: Not equivalent: this formulation critically doesn't include the calls to `begin` and `end`. (Once again, I really want `noexcept(auto)`.)

include/stl2/detail/range/range.hpp
```diff
@@ -23,9 +23,35 @@
 
 STL2_OPEN_NAMESPACE {
 	namespace ext {
-		// iterator_range? view? simple_iterator_and_sentinel_pair?
-		template <Iterator I, Sentinel<I> S = I>
-		using range = tagged_pair<tag::begin(I), tag::end(S)>;
+		// TODO: replace with subrange from P0789
+		template <ext::DestructibleObject I, ext::DestructibleObject S = I>
```
CaseyCarter: This is in `ext`, no need to qualify non-functions.

include/stl2/detail/range/range.hpp
```diff
@@ -23,9 +23,35 @@
...
+
+		template <Iterator I, Sentinel<I> S>
+		struct range<I, S>
+		: tagged_pair<tag::begin(I), tag::end(S)> {
+			range() = default;
+			using tagged_pair<tag::begin(I), tag::end(S)>::tagged_pair;
+		};
+
+		template <Iterator I, Sentinel<I> S>
+		struct range<dangling<I>, dangling<S>>
```
CaseyCarter: Why prefer `range<dangling<I>, dangling<S>>` to `dangling<range<I, S>>`? Specifically, I note that the former is not a range, which I find bothersome.

test/range_access.cpp
```diff
@@ -135,6 +136,9 @@ void test() {
 		static_assert(models::Same<const int*, decltype(__stl2::begin(__stl2::declval<T&>()))>);
 		static_assert(models::Same<const int*, decltype(__stl2::begin(__stl2::declval<const T&>()))>);
 	}
+
+	static_assert(can_begin<__stl2::ext::range<int*,int*>&>);
```
CaseyCarter: s/,(\S)/, \1/g

include/stl2/detail/range/access.hpp
```diff
@@ -321,66 +352,69 @@ STL2_OPEN_NAMESPACE {
...
-			requires(const R& r) {
-				requires SizedSentinel<decltype(__stl2::end(r)), decltype(__stl2::begin(r))>;
-			}
+		requires SizedSentinel<__end_t<const R&>, __begin_t<const R&>>
 		constexpr bool has_difference<R> = true;
 
 		struct fn {
 			template <class T, std::size_t N>
-			constexpr std::size_t operator()(T(&)[N]) const noexcept {
+			constexpr std::size_t operator()(const T(&)[N]) const noexcept {
```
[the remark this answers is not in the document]
ericniebler: For consistency with the other overloads, all of which take `const`-qualified lvalue reference types.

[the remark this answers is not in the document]
CaseyCarter: Which reminds me we haven't implemented ericniebler/stl2#396 yet =(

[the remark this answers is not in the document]
ericniebler: I'm noticing that the resolution of ericniebler/stl2#396 requires the argument of `ranges::size` and `ranges::empty` to be an lvalue. Why shouldn't rvalues be permitted?

ericniebler:
Superseded by #136

PR #138, opened 2018-03-05
Subrange converting ctors cannot be move/copy constructors

ericniebler:


include/stl2/detail/concepts/core.hpp
```diff
@@ -69,6 +61,9 @@ STL2_OPEN_NAMESPACE {
 	template <class T, class... Args>
 	concept bool _OneOf = models::_OneOf<T, Args...>;
 
+	template <class T, class U>
+	concept bool _NotSameAs = __bool<!__is_same_as(__uncvref<T>, __uncvref<U>)>;
```
CaseyCarter: Ditto weird `__bool`.

include/stl2/view/subrange.hpp
```diff
@@ -109,45 +110,47 @@ STL2_OPEN_NAMESPACE {
 			}
 
 			// Not to spec
-			template <Range R>
-			requires !StoreSize && Iterator<safe_iterator_t<R>> &&
+			template <_NotSameAs<subrange> R>
+			requires Range<R> && Iterator<safe_iterator_t<R>> &&
 				ConvertibleTo<iterator_t<R>, I> && ConvertibleTo<sentinel_t<R>, S>
-			constexpr subrange(R&& r)
+			constexpr subrange(R&& r) requires (!StoreSize)
```
CaseyCarter: Ewww, leading + trailing requires clauses. I assume the intent is to draw attention to the similar and dissimilar requirement sets of the two overloads? I think I buy that as a reason to use both leading and trailing `requires`.

ericniebler: For all the members of `subrange` the "sized" constraints are in trailing requires clauses. Makes it a little easier for me to grok if these aren't mixed in with the requirements.

include/stl2/detail/concepts/core.hpp
```diff
@@ -38,23 +38,15 @@ STL2_OPEN_NAMESPACE {
 
 	///////////////////////////////////////////////////////////////////////////
 	// Same [concepts.lib.corelang.same]
-	// Extension: variadic.
 	//
+	template <class T, class U>
+	concept bool Same = __bool<__is_same_as(T, U)> && __bool<__is_same_as(U, T)>;
```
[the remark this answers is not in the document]
ericniebler: For the same reason as for anything else. If `__is_same_as(T, U)` appears elsewhere in someone's requires clauses, I don't want accidental subsumption. I guess your point is that a double-underscore identifier can't appear in a well-formed user program, so why bother, is that right?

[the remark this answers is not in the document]
CaseyCarter: Yes, one reserved identifier is enough to defend against Murphy.

include/stl2/detail/concepts/core.hpp
CaseyCarter: Doubly so given that we don't need to defend internal concepts against subsumption.

include/stl2/view/subrange.hpp
CaseyCarter: I usually try to put non-dependent clauses first, in the hopes that it has (or will have) performance impact. That's probably just wishful thinking on my part.

CaseyCarter: (I've already approved BTW, so don't feel like you *must* address any of these mostly-conversational "issues".)

PR #139, opened 2018-03-07
bring ViewableRange and view::all back into alignment

ericniebler:


ericniebler:
Ok, I'm don't futzing with this.

include/stl2/detail/range/concepts.hpp
```diff
@@ -265,7 +265,7 @@ STL2_OPEN_NAMESPACE {
 		template<class Rng>
 		concept bool ViewableRange =
 			Range<Rng> &&
-			(std::is_lvalue_reference_v<Rng> || View<__f<Rng>>);
+			(requires{typename __begin_t<Rng>;} || View<__f<Rng>>);
```
CaseyCarter: `__begin_t` should be `__end_t` here, or require `Sentinel<__end_t<Rng>, __begin_t<Rng>>` directly.

include/stl2/view/all.hpp
```diff
@@ -29,6 +30,13 @@ STL2_OPEN_NAMESPACE {
 				noexcept(std::is_nothrow_constructible_v<__f<Rng>, Rng>)
 				{ return std::forward<Rng>(rng); }
 
+				template <Range Rng>
+				requires !View<__uncvref<Rng>> &&
+					!std::is_reference_v<Rng> &&
+					requires {typename __begin_t<Rng>;}
```
CaseyCarter: Ditto.

ericniebler:
OK now?

PR #143, opened 2018-04-14
Added `stl2/detail/algorithm/is_permutation.hpp` to `stl2/algorithm.hpp`

cjdb:
Apparently `is_permutation` wasn't found when I included `<experimental/ranges/algorithm>`.

### DIFF
--- a/include/stl2/algorithm.hpp
+++ b/include/stl2/algorithm.hpp
@@ -42,6 +42,7 @@
 #include <stl2/detail/algorithm/is_heap.hpp>
 #include <stl2/detail/algorithm/is_heap_until.hpp>
 #include <stl2/detail/algorithm/is_partitioned.hpp>
+#include <stl2/detail/algorithm/is_permutation.hpp>
 #include <stl2/detail/algorithm/is_sorted.hpp>
 #include <stl2/detail/algorithm/is_sorted_until.hpp>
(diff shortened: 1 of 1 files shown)


cjdb:
This code apparently tries to invoke a deprecated overload:

```cpp
auto v1 = std::vector<T1>{};
auto v2 = std::vector<T2>{};
ranges::is_permutation(v1, v2, std::equal<>{}, proj1, proj2);
```

include/stl2/detail/algorithm/is_permutation.hpp
```diff
@@ -192,8 +192,9 @@ STL2_OPEN_NAMESPACE {
 		}
 		return __stl2::is_permutation(
 			__stl2::begin(rng1), __stl2::end(rng1),
-			__stl2::begin(rng2), std::ref(pred),
-			std::ref(proj1), std::ref(proj2));
+			__stl2::begin(rng2), __stl2::end(rng2),
+			std::ref(pred), std::ref(proj1),
+			std::ref(proj2));
```
CaseyCarter: This change removes the optimization which was the original motivation for having this `SizedRange` overload. It would be better to yank the body out of the deprecated overload on line 72, give it an internal-only name, and have both this overload and line 72 forward into that internal-only backend.

CaseyCarter:
Superseded by #178.

PR #144, opened 2018-04-27
Changes to memory specialisations, as per discussion on cpplang.slack

cjdb:
I think I covered everything.

CC @timsong-cpp

### DIFF
--- a/include/stl2/detail/memory/concepts.hpp
+++ b/include/stl2/detail/memory/concepts.hpp
@@ -19,29 +19,39 @@
 
 STL2_OPEN_NAMESPACE {
 	///////////////////////////////////////////////////////////////////////////
+	// __NoThrowInputIterator [Exposition]
+	//
+	template <class I>
+	concept bool __NoThrowInputIterator =
+		InputIterator<I> &&
+		_Is<reference_t<I>, std::is_lvalue_reference> &&
+		Same<__uncvref<reference_t<I>>, value_type_t<I>>;
+		// Axiom: no exceptions are thrown from increment, assignment, comparison,
+		//        indirection through valid iterators, or comparisons with valid sentinels.
+
+	///////////////////////////////////////////////////////////////////////////
+	// __NoThrowInputRange [Exposition]
+	//
+	template <class Rng>
+	concept bool __NoThrowInputRange =
+		Range<Rng> &&
+		__NoThrowInputIterator<iterator_t<Rng>>;
+
+	///////////////////////////////////////////////////////////////////////////
 	// __NoThrowForwardIterator [Exposition]
 	//
 	template <class I>
 	concept bool __NoThrowForwardIterator =
+		__NoThrowInputIterator<I> &&
 		ForwardIterator<I>;
-		// Axiom: no exceptions are thrown from increment, assignment, comparison,
-		//        or indirection through valid iterators.
 
 	///////////////////////////////////////////////////////////////////////////
-	// __NoThrowForwardIterator [Exposition]
+	// __NoThrowForwardRange [Exposition]
 	//
 	template <class Rng>
 	concept bool __NoThrowForwardRange =
-		Range<Rng> && __NoThrowForwardIterator<iterator_t<Rng>>;
-
-	///////////////////////////////////////////////////////////////////////////
-	// __ReferenceTo [Exposition]
-	//
-	template <class I, class T>
-	concept bool __ReferenceTo = // exposition only
-		InputIterator<I> &&
-		_Is<reference_t<I>, std::is_lvalue_reference> &&
-		Same<__uncvref<reference_t<I>>, T>;
+		__NoThrowInputRange<Rng> &&
+		ForwardRange<Rng>;
 } STL2_CLOSE_NAMESPACE
 
 #endif // STL2_DETAIL_MEMORY_CONCEPTS_HPP
--- a/include/stl2/detail/memory/construct_at.hpp
+++ b/include/stl2/detail/memory/construct_at.hpp
@@ -22,14 +22,14 @@ STL2_OPEN_NAMESPACE {
 	requires Constructible<T, Args...>
 	void __construct_at(T& t, Args&&... args) {
 		::new(const_cast<void*>(static_cast<const volatile void*>(detail::addressof(t))))
-			T(std::forward<Args>(args)...);
+			remove_reference_t<T>(std::forward<Args>(args)...);
 	}
 
 	template <DefaultConstructible T>
 	void __default_construct_at(T& t)
 	{
 		::new(const_cast<void*>(static_cast<const volatile void*>(detail::addressof(t))))
-			T;
+			remove_reference_t<T>;
 	}
 } STL2_CLOSE_NAMESPACE
 
--- a/include/stl2/detail/memory/destroy.hpp
+++ b/include/stl2/detail/memory/destroy.hpp
@@ -36,21 +36,11 @@ STL2_OPEN_NAMESPACE {
 	}
 
 	///////////////////////////////////////////////////////////////////////////
-	// destroy_at [Extension]
-	//
-	template <Destructible T>
-	void destroy_at(T& p) noexcept
-	{
-		p.~T();
-	}
-
-	///////////////////////////////////////////////////////////////////////////
 	// destroy [Extension]
 	//
-	template <InputIterator I, Sentinel<I> S>
+	template <__NoThrowInputIterator I, Sentinel<I> S>
 	requires
-		Destructible<value_type_t<I>> &&
-		__ReferenceTo<I, value_type_t<I>>
+		Destructible<value_type_t<I>>
 	I destroy(I first, S last) noexcept
 	{
 		for (; first != last; ++first) {
@@ -60,10 +50,9 @@ STL2_OPEN_NAMESPACE {
 		return first;
 	}
 
-	template <InputRange Rng>
+	template <__NoThrowInputRange Rng>
 	requires
-		Destructible<value_type_t<iterator_t<Rng>>> &&
-		__ReferenceTo<iterator_t<Rng>, value_type_t<iterator_t<Rng>>>
+		Destructible<value_type_t<iterator_t<Rng>>>
 	safe_iterator_t<Rng> destroy(Rng&& rng) noexcept
 	{
 		return __stl2::destroy(__stl2::begin(rng), __stl2::end(rng));
@@ -72,10 +61,9 @@ STL2_OPEN_NAMESPACE {
 	///////////////////////////////////////////////////////////////////////////
 	// destroy_n [Extension]
 	//
-	template <InputIterator I>
+	template <__NoThrowInputIterator I>
 	requires
-		Destructible<value_type_t<I>> &&
-		__ReferenceTo<I, value_type_t<I>>
+		Destructible<value_type_t<I>>
 	I destroy_n(I first, difference_type_t<I> n) noexcept
 	{
 		return __stl2::destroy(__stl2::make_counted_iterator(std::move(first), n),
--- a/include/stl2/detail/memory/uninitialized_copy.hpp
+++ b/include/stl2/detail/memory/uninitialized_copy.hpp
@@ -28,8 +28,7 @@ STL2_OPEN_NAMESPACE {
 	//
 	template <InputIterator I, Sentinel<I> S, __NoThrowForwardIterator O>
 	requires
-		Constructible<value_type_t<O>, reference_t<I>> &&
-		__ReferenceTo<O, value_type_t<O>>
+		Constructible<value_type_t<O>, reference_t<I>>
 	tagged_pair<tag::in(I), tag::out(O)>
 	uninitialized_copy(I first, S last, O result)
 	{
@@ -41,10 +40,12 @@ STL2_OPEN_NAMESPACE {
 		return {std::move(first), std::move(result)};
 	}
 
+	///////////////////////////////////////////////////////////////////////////
+	// uninitialized_copy [Extension]
+	//
 	template <InputRange Rng, __NoThrowForwardIterator O>
 	requires
-		Constructible<value_type_t<O>, reference_t<iterator_t<Rng>>> &&
-		__ReferenceTo<O, value_type_t<O>>
+		Constructible<value_type_t<O>, reference_t<iterator_t<Rng>>>
 	tagged_pair<tag::in(safe_iterator_t<Rng>), tag::out(O)>
 	uninitialized_copy(Rng&& rng, O result)
 	{
@@ -53,12 +54,24 @@ STL2_OPEN_NAMESPACE {
 	}
 
 	///////////////////////////////////////////////////////////////////////////
+	// uninitialized_copy [Extension]
+	//
+	template <InputRange IRng, __NoThrowForwardRange ORng>
+	requires
+		Constructible<value_type_t<iterator_t<ORng>>, reference_t<iterator_t<IRng>>>
+	tagged_pair<tag::in(safe_iterator_t<IRng>), tag::out(iterator_t<ORng>)>
+	uninitialized_copy(IRng&& irng, ORng& orng)
+	{
+		return __stl2::uninitialized_copy(
+			__stl2::begin(irng), __stl2::end(irng), __stl2::begin(orng));
+	}
+
+	///////////////////////////////////////////////////////////////////////////
 	// uninitialized_copy_n [Extension]
 	//
 	template <InputIterator I, __NoThrowForwardIterator O>
 	requires
-		Constructible<value_type_t<O>, reference_t<I>> &&
-		__ReferenceTo<O, value_type_t<O>>
+		Constructible<value_type_t<O>, reference_t<I>>
 	tagged_pair<tag::in(I), tag::out(O)>
 	uninitialized_copy_n(I first, difference_type_t<I> n, O out)
 	{
--- a/include/stl2/detail/memory/uninitialized_move.hpp
+++ b/include/stl2/detail/memory/uninitialized_move.hpp
@@ -28,8 +28,7 @@ STL2_OPEN_NAMESPACE {
 	//
 	template <InputIterator I, Sentinel<I> S, __NoThrowForwardIterator O>
 	requires
-		Constructible<value_type_t<O>, rvalue_reference_t<I>> &&
-		__ReferenceTo<I, value_type_t<O>>
+		Constructible<value_type_t<O>, rvalue_reference_t<I>>
 	tagged_pair<tag::in(I), tag::out(O)> uninitialized_move(I first, S last, O result)
 	{
 		auto guard = detail::destroy_guard<O>{result};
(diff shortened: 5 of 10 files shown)


include/stl2/detail/memory/concepts.hpp
```diff
@@ -19,29 +19,39 @@
...
 	template <class I>
-	concept bool __NoThrowForwardIterator =
-		ForwardIterator<I>;
+	concept bool __NoThrowInputIterator =
+		InputIterator<I> &&
+		_Is<reference_t<I>, std::is_lvalue_reference> &&
+		Same<__uncvref<reference_t<I>>, value_type_t<I>>;
 		// Axiom: no exceptions are thrown from increment, assignment, comparison,
-		//        or indirection through valid iterators.
+		//        indirection through valid iterators, or comparisons with valid sentinels.
```
timsong-cpp: also move and copy, because you are returning this stuff.

include/stl2/detail/memory/construct_at.hpp
```diff
@@ -22,14 +22,14 @@ STL2_OPEN_NAMESPACE {
 	requires Constructible<T, Args...>
 	void __construct_at(T& t, Args&&... args) {
 		::new(const_cast<void*>(static_cast<const volatile void*>(detail::addressof(t))))
-			T(std::forward<Args>(args)...);
+			remove_reference_t<T>(std::forward<Args>(args)...);
```
timsong-cpp: Not necessary. It looks like the `value_type_t`-based specification in the draft paper doesn't match the actual implementation.

include/stl2/detail/memory/construct_at.hpp
```diff
@@ -22,14 +22,14 @@ STL2_OPEN_NAMESPACE {
...
 	}
 
 	template <DefaultConstructible T>
 	void __default_construct_at(T& t)
 	{
 		::new(const_cast<void*>(static_cast<const volatile void*>(detail::addressof(t))))
-			T;
+			remove_reference_t<T>;
```
timsong-cpp: Likewise.

include/stl2/detail/memory/destroy.hpp
```diff
@@ -60,10 +50,9 @@ STL2_OPEN_NAMESPACE {
 		return first;
 	}
 
-	template <InputRange Rng>
+	template <__NoThrowInputRange Rng>
 	requires
-		Destructible<value_type_t<iterator_t<Rng>>> &&
-		__ReferenceTo<iterator_t<Rng>, value_type_t<iterator_t<Rng>>>
+		Destructible<value_type_t<iterator_t<Rng>>>
 	safe_iterator_t<Rng> destroy(Rng&& rng) noexcept
```
timsong-cpp: Need to document that `begin`/`end` are required to be nothrow somewhere.

include/stl2/detail/memory/uninitialized_copy.hpp
```diff
@@ -41,24 +40,38 @@ STL2_OPEN_NAMESPACE {
...
 	}
 
+	///////////////////////////////////////////////////////////////////////////
+	// uninitialized_copy [Extension]
+	//
+	template <InputRange IRng, __NoThrowForwardRange ORng>
+	requires
+		Constructible<value_type_t<iterator_t<ORng>>, reference_t<iterator_t<IRng>>>
+	tagged_pair<tag::in(safe_iterator_t<IRng>), tag::out(iterator_t<ORng>)>
+	uninitialized_copy(IRng&& irng, ORng& orng)
```
timsong-cpp: It seems pretty safe to `uninitialized_copy` into an rvalue `span`.

include/stl2/detail/memory/uninitialized_move.hpp
```diff
@@ -45,22 +44,33 @@ STL2_OPEN_NAMESPACE {
...
 	}
 
+	///////////////////////////////////////////////////////////////////////////
+	// uninitialized_move [Extension]
+	//
+	template <InputRange IRng, __NoThrowForwardRange ORng>
+	requires
+		Constructible<value_type_t<iterator_t<ORng>>, rvalue_reference_t<iterator_t<IRng>>>
+	tagged_pair<tag::in(safe_iterator_t<IRng>), tag::out(iterator_t<ORng>)>
+	uninitialized_move(IRng&& irng, ORng& orng)
```
timsong-cpp: Likewise.

include/stl2/detail/memory/uninitialized_copy.hpp
```diff
@@ -41,24 +40,38 @@ STL2_OPEN_NAMESPACE {
...
+	{
+		return __stl2::uninitialized_copy(
+			__stl2::begin(irng), __stl2::end(irng), __stl2::begin(orng));
```
timsong-cpp: We should decide what to do for a range size mismatch. Being able to handle that safely is a major advantage of taking two ranges rather than range-and-a-half.

include/stl2/detail/memory/uninitialized_move.hpp
```diff
@@ -45,22 +44,33 @@ STL2_OPEN_NAMESPACE {
...
+	{
+		return __stl2::uninitialized_move(
+			__stl2::begin(irng), __stl2::end(irng), __stl2::begin(orng));
```
timsong-cpp: Ditto.

test/memory/uninitialized_copy.cpp
```diff
@@ -43,6 +43,8 @@ namespace {
 		test(ranges::uninitialized_copy(control.cbegin(), control.cend(), independent.cbegin()));
 		test(ranges::uninitialized_copy(control, independent.begin()));
 		test(ranges::uninitialized_copy(control, independent.cbegin()));
+      test(ranges::uninitialized_copy(control, independent));
```
timsong-cpp: whitespace looks odd (tabs vs spaces?)

include/stl2/detail/memory/uninitialized_copy.hpp
cjdb: How likely are contracts to hit C++20? I'm in favour of `[[expects: ranges::size(irng) <= ranges::size(orng)]]` over throwing for these algorithms in particular.

test/memory/uninitialized_copy.cpp
cjdb: My editor is set to three spaces, and I manually edit them to tabs; looks like this one slipped through the cracks. I'm looking forward to clang-format supporting concepts.

include/stl2/detail/memory/destroy.hpp
cjdb: I'm torn between requiring this as a part of `__NoThrowInputRange` and requiring it right here. I'd prefer it be in the concept, but I'm concerned that will over-constrain `uninitialised_foo`.

include/stl2/detail/memory/uninitialized_copy.hpp
cjdb: I can see benefits for lvalue `span` too.

include/stl2/detail/memory/concepts.hpp
```diff
@@ -18,30 +18,41 @@
...
 STL2_OPEN_NAMESPACE {
+	///////////////////////////////////////////////////////////////////////////
+	// __NoThrowInputIterator [Exposition]
+	//
+	template <class I>
+	concept bool __NoThrowInputIterator =
+		InputIterator<I> &&
+		_Is<reference_t<I>, std::is_lvalue_reference> &&
+		Same<__uncvref<reference_t<I>>, value_type_t<I>>;
+		// Axiom: no exceptions are thrown from increment, copy, move, assignment, comparison,
```
CaseyCarter: "comparison" is weird here, since `I` isn't required to have any comparisons.

include/stl2/detail/memory/concepts.hpp
```diff
@@ -18,30 +18,41 @@
...
+		//        indirection through valid iterators, or comparisons with valid sentinels.
```
CaseyCarter: "comparisons with valid sentinels" is similarly odd since (a) there's no sentinel in sight, and (b) we don't really care if comparison with some value of some type 'S' for which `Sentinel<S,I>` is satisfied can throw if that value and/or type are never passed to one of the algorithms.

Maybe this concept should include the sentinel a la:
```c++
template <class S, class I>
concept bool __NoThrowInputSentinel =
    Sentinel<S, I> && InputIterator<I> && /* ... */;
```

include/stl2/detail/memory/destroy.hpp
```diff
@@ -35,22 +35,12 @@ STL2_OPEN_NAMESPACE {
 		p->~T();
 	}
 
-	///////////////////////////////////////////////////////////////////////////
-	// destroy_at [Extension]
-	//
-	template <Destructible T>
-	void destroy_at(T& p) noexcept
```
CaseyCarter: We'll miss you, `destroy_at(reference)`. 😢

include/stl2/detail/memory/uninitialized_copy.hpp
```diff
@@ -41,24 +41,39 @@ STL2_OPEN_NAMESPACE {
...
+	///////////////////////////////////////////////////////////////////////////
+	// uninitialized_copy [Extension]
+	//
+	template <InputRange IRng, class T>
+	requires
+		Constructible<T, reference_t<iterator_t<IRng>>>
+	tagged_pair<tag::in(safe_iterator_t<IRng>), tag::out(iterator_t<ext::span<T>>)>
+	uninitialized_copy(IRng&& irng, ext::span<T> orng)
+	{
+		STL2_EXPECT(__stl2::size(irng) == __stl2::size(orng));
```
CaseyCarter: `size(InputRange)` isn't required to be valid: this algorithm is underconstrained.

include/stl2/detail/memory/uninitialized_copy.hpp
```diff
@@ -41,24 +41,39 @@ STL2_OPEN_NAMESPACE {
...
 	}
 
+	///////////////////////////////////////////////////////////////////////////
+	// uninitialized_copy [Extension]
+	//
+	template <InputRange IRng, class T>
+	requires
+		Constructible<T, reference_t<iterator_t<IRng>>>
+	tagged_pair<tag::in(safe_iterator_t<IRng>), tag::out(iterator_t<ext::span<T>>)>
+	uninitialized_copy(IRng&& irng, ext::span<T> orng)
```
CaseyCarter: Just `span`? Seems a bit undergeneric.

include/stl2/detail/memory/uninitialized_move.hpp
```diff
@@ -45,22 +45,34 @@ STL2_OPEN_NAMESPACE {
...
 	}
 
+	///////////////////////////////////////////////////////////////////////////
+	// uninitialized_move [Extension]
+	//
+	template <InputRange IRng, class T>
+	requires
+		Constructible<T, rvalue_reference_t<iterator_t<IRng>>>
+	tagged_pair<tag::in(safe_iterator_t<IRng>), tag::out(iterator_t<ext::span<T>>)>
+	uninitialized_move(IRng&& irng, ext::span<T> orng)
```
CaseyCarter: Ditto comments from `uninitialized_copy`.

include/stl2/detail/memory/uninitialized_copy.hpp
CaseyCarter: How about:
```c++
if constexpr(SizedRange<IRng>) {
  STL2_EXPECT(/*...*/);
}
```

include/stl2/detail/memory/uninitialized_copy.hpp
CaseyCarter: I agree that precondition is best here, despite that it can't always be checked because `irng` is single-pass. (You should use `distance` instead of `size` in the precondition.) Throwing is for failure to meet postconditions due to unforeseen (exceptional) circumstances, not for communicating "You gave me garbage input" to the caller.

include/stl2/detail/memory/uninitialized_copy.hpp
cjdb: Not only this, but I think there's an overconstraint on the expectation (should be `size(irng) <= size(orng)` rather than `==`).

include/stl2/detail/memory/uninitialized_copy.hpp
timsong-cpp: ...I meant to use `span` as an example for plausible rvalue out ranges, not literally `span` and `span` alone....

cjdb: That explains a lot...

include/stl2/detail/memory/uninitialized_copy.hpp
timsong-cpp: I'm not particularly convinced that a smaller output range necessarily constitutes garbage input. "Copy until either range is exhausted" seems to be a plausible behavior to me.

include/stl2/detail/memory/uninitialized_copy.hpp
cjdb: My only concern with this is that it invites conflict between size types that are unsigned vs signed (e.g. vector and span).

cjdb: My only concern with this is that it invites conflict between size types that are unsigned vs signed (e.g. vector and span). I've been sitting on this for the past few hours, but haven't come up with a good solution yet.

include/stl2/detail/memory/uninitialized_copy.hpp
cjdb: Added bonus: this solves the signed vs unsigned problem.

include/stl2/detail/memory/uninitialized_copy.hpp
CaseyCarter: Use `distance` - which is always signed - instead of `size`?

include/stl2/detail/memory/uninitialized_copy.hpp
CaseyCarter: > "Copy until either range is exhausted" seems to be a plausible behavior to me.

I can see arguments for any of:
1. "Copy until either range is exhausted" (no precondition on range lengths)
2. "Always copy every element of the input" (|input| <= |output|)
3. "Always write to every element of the output" (|input| >= |output|)
4. "Always copy each element of the input into each element of the output" (|input| == |output|)

(2) most closely matches the behavior of output iterator algorithms - they simply assume that the output range has enough space to copy every input element. We need a clear benefit to disregard consistency of `copy(irange, orange)` with `copy(first, last, result)`.

I think the "clear benefit" of choosing (1) is that a caller can more easily adapt an algorithm to each of the other semantics than if we were to choose (2), (3), or (4). Implementing (1) atop (4), for example, is just barely feasible when `irange` and `orange` are both single-pass - certainly not a job for someone who isn't an expert in range adaptors. Implementing any of the others atop (1), however, is as simple as comparing the iterators returned from the algorithm with the end sentinels of the ranges and `assert`ing.

I'm a bit nervous making this level of policy decision in a code review without getting wider community input - realistically we're talking about more than just `uninitialized_copy`. @ericniebler, thoughts?

include/stl2/detail/memory/uninitialized_copy.hpp
cjdb: Great idea (and I'll be using `distance` from now on, forever), but I'm keen on following the discussion above in favour of this.

cjdb:
@CaseyCarter The Travis check is failing for reasons that are probably unrelated to this MR.

include/stl2/detail/memory/uninitialized_copy.hpp
cjdb: I made an issue here: https://github.com/ericniebler/stl2/issues/556

test/memory/uninitialized_move.cpp
```diff
@@ -16,25 +16,29 @@
...
 #include "common.hpp"
 
 namespace ranges = __stl2;
+using ranges::ext::span;
 
 namespace {
-	template <typename T>
-	bool empty(const Array<T>& a) {
-		return ranges::all_of(a, &T::empty);
+	template <ranges::Range Rng>
```
CaseyCarter: This is underconstrained - should be something like:
```c++
template <ranges::InputRange Rng>
requires requires {
  typename ranges::value_type_t<Rng>;
  &ranges::value_type_t<Rng>::empty;
  requires ranges::Invocable<
    decltype(&ranges::value_type_t<Rng>::empty),
    ranges::reference_t<ranges::iterator_t<const Rng>>>;
}
```

test/memory/uninitialized_move.cpp
```diff
@@ -16,25 +16,29 @@
...
+	bool empty(const Rng& rng, const std::ptrdiff_t n) {
+		return ranges::all_of(ranges::make_counted_iterator(rng.begin(), n), ranges::default_sentinel{},
+			&ranges::value_type_t<Rng>::empty);
 	}
 
-	template <typename T>
-	requires std::is_fundamental<T>::value
-	bool empty(const Array<T>&) {
+	template <ranges::Range Rng>
```
CaseyCarter: Ditto `InputRange`.

test/memory/uninitialized_move.cpp
```diff
@@ -16,25 +16,29 @@
...
+	requires std::is_fundamental<ranges::value_type_t<Rng>>::value
```
CaseyCarter: In product code I'd recommend requiring `typename ranges::value_type_t<Rng>` before using that type in a requirement like this (because sometimes better error messages) but this is test code so *shrug*.

include/stl2/detail/memory/uninitialized_move.hpp
CaseyCarter: Ditto "`ORng&&` and the return type `tagged_pair<tag::in(safe_iterator_t<IRng>), tag::out(safe_iterator_t<ORng>)>`"

test/memory/uninitialized_move.cpp
cjdb: What is the difference between `requires Invocable<...>;` and `Invocable<...>;` here?

cjdb: Ah, I understand this now. Nested requirements are effectively the `typename` of concepts.

CaseyCarter: Inside a requires expression:
* `requires Concept</*...*/>;` is a nested requirement that requires `Concept</*...*/>` to be satisfied. 
* `Concept</*...*/>` is an expression requirement that requires `Concept</*...*/>` to be a valid expression.

It's extremely unfortunate that the syntaxes are so similar: forgetting a `requires` on a nested requirement results in a valid program with totally different meaning.

include/stl2/detail/memory/uninitialized_copy.hpp
```diff
@@ -20,18 +20,18 @@
 #include <stl2/detail/memory/construct_at.hpp>
 #include <stl2/detail/memory/destroy.hpp>
 #include <stl2/detail/range/concepts.hpp>
+#include <stl2/detail/span.hpp>
```
CaseyCarter: This is unnecessary now, right?

CaseyCarter: (and in `uninitialized_move`)

test/memory/uninitialized_move.cpp
```diff
@@ -16,25 +16,39 @@
...
 #include <stl2/view/take_exactly.hpp>
+#include <algorithm>
 #include <cstdint>
 #include <memory>
 #include <vector>
 #include "../simple_test.hpp"
 #include "common.hpp"
 
 namespace ranges = __stl2;
+using ranges::ext::span;
```
CaseyCarter: This test, however *does* need `span.hpp`.

CaseyCarter:
(Looks at Travis) Oh,  you *did* rebase onto master. One of your tests is segfaulting.

cjdb:
I _think_ I fixed that segfault (I can't seem to replicate it: are you using any special flags on Travis?). Turns out that I was using `independent` as an input range, but it has its contents destroyed after each test, so that was a mistake.

cjdb:
Finally. FYI https://twitter.com/cjdb_ns/status/990868731768303616

PR #145, opened 2018-04-29
Implement P0970 for span and basic_string_view

CaseyCarter:
Also relocate `span` test into `test/view`.

### DIFF
--- a/test/range_access.cpp
+++ b/test/range_access.cpp
@@ -11,13 +11,16 @@
 // Project home: https://github.com/caseycarter/cmcstl2
 //
 #include <stl2/iterator.hpp>
+#include <stl2/detail/algorithm/find.hpp>
 #include <iostream>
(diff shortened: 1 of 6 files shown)


test/range_access.cpp
```diff
@@ -11,13 +11,16 @@
 // Project home: https://github.com/caseycarter/cmcstl2
 //
 #include <stl2/iterator.hpp>
+#include <stl2/detail/algorithm/find.hpp>
 #include <iostream>
 #include <vector>
 #include "simple_test.hpp"
 
+namespace ranges = __stl2;
```
CaseyCarter: There are a ton of non-functional changes in this file replacing uses of `__stl2` with `ranges` and indenting namespace contents. The actual function change is the addition of `test_string_view_p0970` on line 172.

PR #148, opened 2018-05-06
Fixed files that didn't compile on GCC 8 and added Travis support

cjdb:
For whatever reason, the deleted code doesn't work with GCC 8. I suspect a compiler bug, since both `decltype(auto)` and `auto` pass for return types.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,19 @@ matrix:
       os: linux
       addons: *gcc7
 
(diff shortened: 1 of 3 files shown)


CaseyCarter:
Thanks for the report. I narrowed the issue down to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85706, and worked around a bit differently in df41efe31e3b03ad5e93231af8e86794d340072b to preserve the effect of the expression SFINAE on the return type.

Issue #149, opened 2018-05-18
istreambuf_iterator constructed with istream at eof isn't detected


cjdb:
When comparing a freshly-constructed `istreambuf_iterator` with an `istream` that has already reached EOF, `istreambuf_iterator != default_sentinel` returns `true`. This behaviour differs from `std::istreambuf_iterator`.

```cpp
#include <experimental/ranges/iterator>
#include <iostream>
#include <sstream>

int main()
{
   auto in = std::istringstream{"hello"};
   for (auto i = __stl2::istreambuf_iterator<char>{in}; i != __stl2::default_sentinel{}; ++i) {
      std::cout << *i << '\n';
   }
   for (auto i = __stl2::istreambuf_iterator<char>{in}; i != __stl2::default_sentinel{}; ++i) {
      std::cout << *i << '\n';
   }
}
```

Interestingly, this is only on the first pass: commenting out line 16 causes the program to complete without error.

CaseyCarter:
> commenting out line 16

I only count 14 lines.

cjdb:
>I only count 14 lines.

Ahh! I took a few lines out from GitHub, but not my editor. Should be line 12.

cjdb:
This might be related to #28.

CaseyCarter:
`istreambuf_iterator` is, unsuprisingly, broken. [[istreambuf.iterator.cons]](http://eel.is/c++draft/istreambuf.iterator.cons):

> 1 For each `istreambuf_­iterator` constructor in this subclause, **an end-of-stream iterator is constructed if and only if the exposition-only member `sbuf_`­ is initialized with a null pointer value**.
> ```c++
> constexpr istreambuf_iterator() noexcept;
> ```
> 2 *Effects:* Initializes `sbuf_­` with `nullptr`.
> ```c++
> istreambuf_iterator(istream_type& s) noexcept;
> ```
> 3 *Effects:* Initializes `sbuf_­` with `s.rdbuf()`.
> ```c++
> istreambuf_iterator(streambuf_type* s) noexcept;
> ```
> 4 *Effects:* Initializes `sbuf_­` with `s`.
> ```c++
> istreambuf_iterator(const proxy& p) noexcept;
> ```
> 5 *Effects:* Initializes `sbuf_­` with `p.sbuf_­`.

[[istreambuf.iterator.ops]](http://eel.is/c++draft/istreambuf.iterator.ops) para 5 & 6:

> ```c++
[...]

PR #153, opened 2018-05-19
Patch for istreambuf_iterator == default_sentinel

cjdb:
Before this patch, the test below fails.

```cpp
auto in = std::istringstream{};
assert(istreambuf_iterator{in} == default_sentinel{});
```

Related to #149.


include/stl2/detail/iterator/istreambuf_iterator.hpp
```diff
@@ -165,7 +165,7 @@ STL2_OPEN_NAMESPACE {
 				return at_end() == that.at_end();
 			}
 			STL2_CONSTEXPR_EXT bool equal(default_sentinel) const noexcept {
-				return sbuf_ == nullptr;
+				return cursor{}.equal(*this);
```
CaseyCarter: `return at_end();` would be simpler here.

PR #164, opened 2018-07-05
Simplify take_view:

CaseyCarter:
* Remove unused includes

* We don't need the nested typenames, so get rid of the extra layer of derivation

* The sentinel comparisons don't require iterators to be `EqualityComparable`

* Combine `begin` and `end` overloads with `if constexpr`

* Disable mutable members when `SimpleView<R>`

* When `R` is sized and not random access, use `default_sentinel`


CaseyCarter:
Don't merge this until P0896 is updated.

CaseyCarter:
> Don't merge this until P0896 is updated.

It is now so updated.

PR #165, opened 2018-07-05
Add headers <stl2/ranges.hpp> and <experimental/ranges/ranges>

CaseyCarter:


ericniebler:
Why `<stl2/` ?

CaseyCarter:
All of the `<experimental/ranges/foo>` headers have a corresponding public `<stl2/foo.hpp>` header. It's part of the possibly-ill-considered "This library can coexist alongside an actual Ranges TS implementation" design.

CaseyCarter:
The question you should be asking is "Why `<ranges>` when the TS header is named `<range>`?"

PR #174, opened 2018-07-23
Extension views

cjdb:
Adds the extension views' CPOs to namespace `view::ext`.


include/stl2/view/take_exactly.hpp
```diff
@@ -88,10 +93,31 @@ STL2_OPEN_NAMESPACE {
 			constexpr iter_difference_t<iterator_t<Base>> size() const noexcept { return n_; }
 			constexpr bool empty() const noexcept { return n_ == 0; }
 		};
+
+		template <InputRange R>
```
CaseyCarter: Why `InputRange` here instead of `Range`? I think this view adaptor will work as written with output ranges.

include/stl2/view/take_exactly.hpp
```diff
@@ -88,10 +93,31 @@ STL2_OPEN_NAMESPACE {
...
 	template <class V>
 	struct enable_view<ext::take_exactly_view<V>> : std::true_type {};
+
+	namespace view::ext {
+		struct __take_exactly_fn : detail::__pipeable<__take_exactly_fn> {
+			template <class V>
+			auto operator()(V&& view, iter_difference_t<iterator_t<V>> const n) const
+			STL2_NOEXCEPT_REQUIRES_RETURN
+			(
+				__stl2::ext::take_exactly_view(std::forward<V>(view), n)
```
CaseyCarter: s/b `__stl2::ext::take_exactly_view(all(std::forward<V>(view), n))` to convert `ViewableRange`s to `View`s.

test/view/repeat_n_view.cpp
```diff
@@ -35,6 +35,13 @@ int main() {
 		auto v = ranges::ext::repeat_n_view<empty>{{}, (1ULL << 20)};
 		static_assert(sizeof(decltype(v.begin())) == sizeof(std::ptrdiff_t));
 	}
+	{
+		auto v = ranges::view::ext::repeat_n(9, 10);
+		static_assert(ranges::View<decltype(v)>);
+		static_assert(ranges::RandomAccessIterator<decltype(v.begin())>);
+		// static_assert(ranges::SizedView<decltype(v)>);
```
CaseyCarter: Uncomment this and change `SizedView` to `SizedRange`?

test/view/repeat_view.cpp
```diff
@@ -97,6 +98,13 @@ int main() {
 				const std::vector<int>*>);
 		CHECK(first->size() == N);
 	}
+	{
+		auto v = ranges::view::ext::repeat(9) | ranges::view::take(10);
+		static_assert(ranges::View<decltype(v)>);
+		static_assert(ranges::RandomAccessIterator<decltype(v.begin())>);
+		// static_assert(ranges::SizedView<decltype(v)>);
```
CaseyCarter: Ditto `SizedView` => `SizedRange`?

include/stl2/view/take_exactly.hpp
cjdb: This one mimics `take_view`'s deduction guide. Should I update that too?

CaseyCarter: Yes, please.

include/stl2/view/take_exactly.hpp
cjdb: Ibid

include/stl2/view/take_exactly.hpp
CaseyCarter: ...and I'll make a note to relax `take_view` similarly in P0896.

test/view/repeat_view.cpp
cjdb: I've been thinking about what `repeat_view::size` should return. Even if we bring the `cardinality` type over from range-v3, it feels like an arbitrary result (IIUC the size is -3).

CaseyCarter: `repeat_view` should have no member `size` - it's not finite, so it has no well-defined size.  (And yes, it didn't occur to me that we don't have range-v3's cardinality machinery so `take_view<repeat_view<Foo>>` will necessarily not be a sized range.)

range-v3's `cardinality` isn't strictly a size: values `>= 0` are sizes, values `< 0`:
* `unknown` is potentially infinite, (e.g., a `filter_view` adapting an `infinite` underlying range)
* `finite` is not known statically but definitely finite (e.g. `vector`)
* `infinite` is definitely infinite. (e.g. `repeat_view`)

CaseyCarter: In other words: ignore this suggestion.

PR #175, opened 2018-07-28
Updated repeat and repeat_n so they permit lvalues

cjdb:
The salient requirement for repeat and repeat_n is that their T is a copy-constructible object, but when an lvalue is passed to either, it was forwarded, and so was treated as an lvalue reference, which breaks that requirement.

This fix separates lvalue references from rvalue references, rather than conflating them using forwarding references.

### DIFF
--- a/include/stl2/view/repeat.hpp
+++ b/include/stl2/view/repeat.hpp
@@ -75,9 +75,15 @@ STL2_OPEN_NAMESPACE {
 	namespace view::ext {
 		class __repeat_fn {
 			template <class T>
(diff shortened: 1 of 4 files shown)


CaseyCarter:
I was going to comment "We should perfect-forward the argument to `repeat` into the `repeat_view`, but then I pushed https://github.com/cjdb/cmcstl2/pull/19 instead.

ericniebler:
I don't really use Visual Studio at all anymore. That would probably change if it were ever able to compile range-v3.

PR #177, opened 2018-07-29
Added view::ext::drop and tests.

cjdb:
Here's an attempt at `drop`, influenced by `take` (but not range-v3's `drop`). Tests are from range-v3. `drop_exactly`, `drop_while`, and `take_while` to come soon.


include/stl2/view/drop.hpp
```diff
@@ -0,0 +1,228 @@
...
+			{
+				return __sentinel<SimpleView<R>>(__stl2::end(base_));
+			}
+
+			constexpr auto end() const
+			requires
+				Range<R const>
+			{
+				return __sentinel<true>(__stl2::end(base_));
+			}
```
cjdb: Both of these `end`s are wrong.

include/stl2/view/drop.hpp
```diff
@@ -0,0 +1,228 @@
...
+				return __stl2::next(__stl2::begin(x.base_), x.count_);
+			}
+
+			template <class X>
+			static constexpr decltype(auto) begin_impl(X&& x)
+			requires
+				SizedRange<R>
+			{
+				D const dist = __stl2::distance(x.base_);
+				return (__stl2::next(__stl2::begin(x.base_), dist < x.count_ ? dist : x.count_));
```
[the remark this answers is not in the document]
CaseyCarter: An `STL2_EXPECT(count_ >= 0);` before the `next` call would likely improve codegen by telling the optimizer that it can eliminate the code that handles negative counts for bidirectional iterators in `next`.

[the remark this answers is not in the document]
CaseyCarter: And now the nasty part: this is `O(1)` for random access ranges, but nothing weaker. For weaker underlying ranges, we need to cache the first iterator in a `detail::non_propagating_cache`. (`reverse_view` is a good example.) Consequently, the `const` overloads of `begin` and `end` need to be enabled only when `R` models `RandomAccessRange`.

include/stl2/view/drop.hpp
```diff
@@ -0,0 +1,228 @@
...
+			{}
+
+			constexpr sentinel_t<Base> base() const
+			{
+				return end_;
+			}
+
+			friend constexpr bool operator==(__sentinel const& x, iterator_t<Base> const& y)
+			requires
+				EqualityComparable<iterator_t<Base>>
```
CaseyCarter: It doesn't matter if we axe this custom sentinel, but FYI, these `EqualityComparable` requirements are bogus: there are no iterators being compared here. I think @ericniebler put them in `take_view` just to see if I was paying attention.

include/stl2/view/drop.hpp
CaseyCarter: There's a "burning the candle at both `end`s" joke here somewhere.

include/stl2/view/take_exactly.hpp
```diff
@@ -39,7 +40,7 @@ STL2_OPEN_NAMESPACE {
 			take_exactly_view() requires DefaultConstructible<Base> = default;
 
 			constexpr take_exactly_view(Base view, iter_difference_t<iterator_t<Base>> n)
-			noexcept(std::is_nothrow_move_constructible<Base>::value)
+			noexcept(noexcept(std::is_nothrow_move_constructible<Base>::value))
```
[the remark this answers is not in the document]
cjdb: Oooh so `noexcept(noexcept(expr))` is only useful iff `expr` is an expression that potentially throws. I did wonder why we always needed to write `noexcept(noexcept(expr))`, and the answer is "you don't".

include/stl2/view/drop.hpp
```diff
@@ -0,0 +1,228 @@
...
+			}
+
+			constexpr auto end() const
+			requires
+				Range<R const>
+			{
+				return __sentinel<true>(__stl2::end(base_));
+			}
+
+			constexpr auto size() const
```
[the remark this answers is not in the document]
CaseyCarter: Ironically enough, `take_view`. Reviewing this change initially confused me - I kept suggesting changes that were vaguely familiar, before realizing I'd already made the same changes to `take_view` but not yet merged them. If you look at `take_view` as of this morning, it does the right thing in `size`.

include/stl2/view/drop.hpp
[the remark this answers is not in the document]
cjdb: I think this and the `!SimpleView` requirement on the non-`const` overload are altogether a bit too much. `std::list` isn't a `RandomAccessRange` and it also isn't a `SimpleView`, but it should be usable with `drop`. I'm partial to making a mutable `non_propagating_cache`, rather than inheriting from it.

test/view/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@
 add_stl2_test(span span span.cpp)
 add_stl2_test(view.common view.common common_view.cpp)
 add_stl2_test(view.counted view.counted counted_view.cpp)
+# add_stl2_test(view.drop_exactly view.drop_exactly drop_exactly_view.cpp)
```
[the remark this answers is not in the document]
cjdb: Would you like me to remove the commented-out tests too? My intention is to enable them once their friends are in cmcstl2.

include/stl2/view/take.hpp
```diff
@@ -40,7 +40,7 @@ STL2_OPEN_NAMESPACE {
 					if constexpr (RandomAccessRange<RR>) {
 						return __stl2::begin(self.base_);
 					} else {
-						return counted_iterator{__stl2::begin(self.base_), self.size()};
+						return counted_iterator{__stl2::begin(self.base_), static_cast<D>(self.size())};
```
cjdb: This was exposed by accident -- I decided to see if a test worked for `take_view` and it triggered an implicit conversion warning.

include/stl2/view/drop.hpp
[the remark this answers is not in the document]
CaseyCarter: The point of the `!SimpleView` constraint on the non-`const` overloads is to avoid instantiating them when they would be equivalent to the `const` overloads. Adding the `RandomAccessRange` constraint to the `const` overloads needs to be reflected in the non-`const` constraint to maintain that equivalence: they should be constrained with `!(SimpleView<R> && RandomAccessRange<R>)`.

Inheriting from `non_propagating_cache` makes it EBO away when it's disabled and therefore empty, which is nice. `mutable` is a recipe for violating [[res.on.data.races]](http://eel.is/c++draft/res.on.data.races) - updating a `mutable` cache from a `const` member function means that simultaneous calls to e.g. `begin` can cause data races.

test/view/CMakeLists.txt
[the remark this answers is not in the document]
CaseyCarter: I hadn't yet seen them when I made this comment, or I would have clarified: the commented-out tests with the descriptive comment that tells us *why* they are commented out are  fine.

include/stl2/view/drop.hpp
```diff
@@ -0,0 +1,117 @@
...
+			static constexpr auto begin_impl(X& x)
+			{
+				if (!x.begin_) {
+					if constexpr (SizedRange<__maybe_const<is_const_v<X>, R>>) {
+						D const dist = __stl2::distance(x.base_);
+						STL2_EXPECT(x.count_ >= 0);
+						x.begin_ = __stl2::next(__stl2::begin(x.base_), dist < x.count_ ? dist : x.count_);
+					}
+					else {
+						x.begin_ = __stl2::next(__stl2::begin(x.base_), x.count_);
```
CaseyCarter: This should call the "bounded" `next` overload `next(begin(base_), count_, end(base_))` to avoid running past the end of the base range.

include/stl2/view/drop.hpp
```diff
@@ -0,0 +1,117 @@
...
+				_ConstructibleFromRange<R, O>
+			constexpr drop_view(O&& o, D count)
+				: base_(view::all(std::forward<O>(o))),
+				  count_(count)
+			{}
+
+			constexpr R base() const { return base_; }
+
+			constexpr auto begin() requires !SimpleView<R> { return begin_impl(*this); }
+			constexpr auto begin() const requires Range<R const>
```
CaseyCarter: Despite that Eric and I are both east `const`-ers, cmcstl2 uses `const` west - at least for occurrences in the interface - to follow the conventions of the IS and minimize differences between the declarations as they appear in the implementation vs. the spec. I can live with the east `const` usage in the function bodies and interfaces of private functions (I hate `const` west enough to live with the inconsistency.)

include/stl2/view/drop.hpp
```diff
@@ -0,0 +1,117 @@
...
+			constexpr auto size() const requires SizedRange<R const> { return size_impl(*this); }
+		private:
+			R base_;
+			D count_;
+			mutable detail::non_propagating_cache<iterator_t<R>> begin_;
+
+			template <class X>
+			static constexpr auto begin_impl(X& x)
+			{
+				if (!x.begin_) {
```
CaseyCarter: Data race: simultaneous calls to `begin() const` from multiple threads generate conflicting reads and writes to `begin_`.

include/stl2/view/drop.hpp
```diff
@@ -47,36 +49,62 @@ STL2_OPEN_NAMESPACE {
...
+			constexpr auto begin()
+			requires !(SimpleView<R> && RandomAccessRange<R>)
+			{ return begin_impl(*this); }
+			constexpr auto begin() const requires Range<R const> && RandomAccessRange<R const>
 			{ return begin_impl(*this); }
 
-			constexpr auto end() requires !SimpleView<R> { return end_impl(*this); }
-			constexpr auto end() const requires Range<R const>
+			constexpr auto end()
+			requires !(SimpleView<R> && RandomAccessRange<R>)
```
CaseyCarter: I suppose the `end` overloads don't need the `RandomAccessRange` constraints, since they don't need to access the cache.

CaseyCarter: Admittedly it could be confusing to have differing requirements on `begin` and `end` - I don't know that I have a preference here.

include/stl2/view/drop.hpp
```diff
@@ -47,36 +49,62 @@ STL2_OPEN_NAMESPACE {
...
+					STL2_EXPECT(x.count_ >= 0);
+					return __stl2::next(__stl2::begin(x.base_), dist < x.count_ ? dist : x.count_);
+				};
+
+				[[maybe_unused]] auto unchecked_begin = [&x]{
+					return __stl2::next(__stl2::begin(x.base_), x.count_);
+				};
+
+				if constexpr (RandomAccessRange<__maybe_const<is_const_v<X>, R>>) {
 					if constexpr (SizedRange<__maybe_const<is_const_v<X>, R>>) {
```
CaseyCarter: You could merge the `XXX_begin` lambdas by moving this `if constexpr` switch inside.

include/stl2/view/drop.hpp
```diff
@@ -47,36 +49,62 @@ STL2_OPEN_NAMESPACE {
...
-						STL2_EXPECT(x.count_ >= 0);
-						x.begin_ = __stl2::next(__stl2::begin(x.base_), dist < x.count_ ? dist : x.count_);
+						return checked_begin();
 					}
 					else {
-						x.begin_ = __stl2::next(__stl2::begin(x.base_), x.count_);
+						return unchecked_begin();
 					}
 				}
+				else {
```
CaseyCarter: Would binding a reference to `x.cached_begin()` here make the rest of this more readable?

include/stl2/view/drop.hpp
CaseyCarter: Aside: I *hate* that GCC and MSVC emit "unused" warnings for things that are only referenced in a discarded branch of an `if constexpr`. I must have put two dozen workarounds in the rework of `<variant>` I was working on this week.

cjdb: I did away with `checked_begin` and `unchecked_begin` when you mentioned that there's actually a bounded `next` (I thought you were saying "here's a good reason to propose one" earlier in the week).

I feel kind of sheepish having only an assert inside an `if constexpr` branch.

```cpp
auto compute_begin = [&x]{
	if constexpr (SizedRange<__maybe_const<is_const_v<X>, R>>) {
		STL2_ASSERT(x.count_ >= 0);
	}
	return __stl2::next(__stl2::begin(x.base_), x.count_, __stl2::end(x.base_));
};
```

On the plus side, no `[[maybe_unused]]` attribute is necessary!

CaseyCarter: Yeah, I like this better. You can push that `ASSERT` up (and make it an `EXPECT` since it's "free", and may help the optimizer in release builds) and then "unbranch" some of the control flow.

CaseyCarter: > Yeah, I like this better.

Oops - except that simply the bounded `next` call is O(n) for `SizedRange`s whose iterator and sentinel don't model `SizedSentinel` - we should probably keep the `if constexpr (SizedRange)` switch inside `compute_begin`.

CaseyCarter: > ... we should probably keep the `if constexpr (SizedRange)` switch inside `compute_begin`

FYI, this is the only blocker now. I think the PR is good otherwise.

include/stl2/view/drop.hpp
```diff
@@ -75,7 +75,13 @@ STL2_OPEN_NAMESPACE {
 			{
 				auto compute_begin = [&x]{
 					STL2_EXPECT(x.count_ >= 0);
-					return __stl2::next(__stl2::begin(x.base_), x.count_, __stl2::end(x.base_));
+					if constexpr (SizedRange<__maybe_const<is_const_v<X>, R>>) {
+						D const dist = __stl2::distance(x.base_);
+						return __stl2::next(__stl2::begin(x.base_), dist < x.count_ ? dist : x.count_);
```
CaseyCarter: Final nitpick and then I swear I'll stop: advancing `begin(rng)` by `distance(rng)` is unnecessarily O(N) in if the range is `Bounded` but not `RandomAccess`; we could use `end(rng)` directly. How about:
```c++
D n = __stl2::distance(x.base_);
if (n > x.count_) {
    n = x.count_;
} else if constexpr (!RandomAccessRange<X> && BoundedRange<X>) {
    return __stl2::end(x.base_);
}
return __stl2::next(__stl2::begin(x.base_), n);
```

...and now I'm thinking that it would be an improvement to decrement `end` if `x.count_ > distance(x.base_) / 2` and the range is `Bounded` and `Bidirectional`:

```c++
D const size = __stl2::distance(x.base_);
D n = x.count_;
constexpr bool BoundedNonRandom = BoundedRange<X> && !RandomAccessRange<X>;
if (n >= size) {
    if constexpr (BoundedNonRandom && !BidirectionalRange<X>) {
        return __stl2::end(x.base_);
    }
    n = size;
}
if constexpr (BoundedNonRandom && BidirectionalRange<X>) {
    if (n > size / 2) {
        return __stl2::prev(__stl2::end(x,base_), size - n));
    }
}
return __stl2::next(__stl2::begin(x.base_), n);
```
There's a *lot* of subtlety here: we ought to wrap up this entire lambda and propose it as `nth_iterator(range, n)` (or `index`, or `at`, or ...)

cjdb: Where `BoundedRange` is `CommonRange`?

cjdb: > There's a lot of subtlety here: Should we wrap up this entire lambda and propose it as `nth_iterator(range, n)` (or `index`, or `at`, or ...)?

IMO iff `rng | drop(n)` isn't quite enough to always say "start from `n`". Otherwise I guess it could be exposition-only for `drop` and friends?

CaseyCarter: Cursed rename - I'm still not used to it. I suspect there will be other use cases for `nth_iterator`; we can propose it if and when they appear and validate my suspicions.

CaseyCarter:
Thanks for all the work that I made you do on this, Chris!

PR #180, opened 2018-08-05
Added take_while_view and drop_while_view as an extension to cmcslt2

cjdb:
As promised, `take_while`, hot on the heels of `drop` (`take_while` was easier for me to mentally model than `drop_while`, but I suspect to drop the latter not long after this one's bugs are ironed out).


cjdb:
And here's `drop_while`. It was simpler than I'd expected (I had to peek at range-v3 to see if `const` overloads for `begin` and `end` were present).

include/stl2/view/take_while.hpp
```diff
@@ -0,0 +1,116 @@
...
+		template <bool Const>
+		class take_while_view<R, Pred>::__sentinel {
+			using Base = __maybe_const<Const, R>;
+			sentinel_t<Base> end_ {};
+			const Pred* pred_;
+		public:
+			__sentinel() = default;
+
+			constexpr explicit __sentinel(sentinel_t<Base> end, const Pred* pred)
+			: end_(end), pred_(std::move(pred)) {}
```
CaseyCarter: You could make an argument for `std::move(end)`, but `std::move(pointer)` is just silly ;) (and on 87.)

include/stl2/view/take_while.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+			constexpr __sentinel(__sentinel<!Const> s)
+			requires Const && ConvertibleTo<sentinel_t<R>, sentinel_t<Base>>
+			: end_(s.base()), pred_(std::move(s.pred_)) {}
```
CaseyCarter: We need to `friend __sentinel<false>;` so it can access `pred_`. (The compiler gives us no help here thanks to https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67225 turning off access control.)

include/stl2/view/take_while.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+	namespace view::ext {
+		struct __take_while_fn : detail::__pipeable<__take_while_fn> {
+			template <class Rng, class Pred>
+			constexpr auto operator()(Rng&& rng, Pred&& pred) const
+			STL2_NOEXCEPT_REQUIRES_RETURN(
+				__stl2::ext::take_while_view{view::all(static_cast<Rng&&>(rng)), std::forward<Pred>(pred)}
+			)
+
+			template <CopyConstructible Pred>
```
CaseyCarter: Ditto needs to require an object type.

include/stl2/view/take_while.hpp
```diff
@@ -0,0 +1,116 @@
...
+			: end_(end), pred_(std::move(pred)) {}
+
+			constexpr __sentinel(__sentinel<!Const> s)
+			requires Const && ConvertibleTo<sentinel_t<R>, sentinel_t<Base>>
+			: end_(s.base()), pred_(std::move(s.pred_)) {}
+
+			constexpr sentinel_t<Base> base() const { return end_; }
+
+			friend constexpr bool operator==(const __sentinel& x, const iterator_t<Base> y)
+			{ return x.end_ == y || !(*x.pred_)(*y); }
```
CaseyCarter: `*x.pred_` is an lvalue `const Pred`, and we didn't require `IndirectUnaryPredicate<const Pred, iterator_t<Base>>`.

test/view/drop_while_view.cpp
```diff
@@ -0,0 +1,87 @@
...
+	//  static int const rgi[] = {0,1,2,3,4,5,6,7,8,9};
+	//  int cnt = 0;
+	//  auto mutable_only = view::drop_while(rgi, [cnt](int) mutable { return ++cnt <= 5;});
+	//  CHECK_EQUAL(mutable_only, {5,6,7,8,9});
+	//  static_assert(View<decltype(mutable_only)>());
+	//  static_assert(!View<decltype(mutable_only) const>());
+
+	// Check with a mutable predicate
+	{
+		auto rng0 = view::iota(10) | view::drop_while([](int i) mutable { return i < 25; });
```
CaseyCarter: You should probably mutate some non-salient state in the lambda just out of testing paranoia - bump a call counter or flip a `bool` or whatever.

test/view/drop_while_view.cpp
```diff
@@ -0,0 +1,87 @@
...
+	//  static_assert(range_cardinality<decltype(rng1)>::value == ranges::finite);
+	static_assert(ranges::BidirectionalRange<decltype(rng1)>);
+	// ::models<concepts::BoundedView>(aux::copy(rng1));
+	static_assert(ranges::BidirectionalIterator<decltype(rng1.begin())>);
+	 CHECK(rng1.begin() == rng1.end());
+
+	/////////////////////////////////////////////////////////////////////////////////////////////////
+	//																															  //
+	// 						DISABLED because the lambda is not equality-preserving.						  //
+	//							Left in the review only to confirm that this is the case.					  //
```
CaseyCarter: You're right - this test is ill-formed with no diagnostic required.

CaseyCarter: Do you think it would be a good idea for debug compiles to call the predicate twice per invocation and assert that the results are the same to flush out bugs like this? I hate to slow down debug compiles, but this is a very easy bug to write.

test/view/drop_while_view.cpp
```diff
@@ -0,0 +1,87 @@
...
+	//  static_assert(View<decltype(mutable_only)>());
+	//  static_assert(!View<decltype(mutable_only) const>());
+
+	// Check with a mutable predicate
+	{
+		auto rng0 = view::iota(10) | view::drop_while([](int i) mutable { return i < 25; });
+		//  static_assert(range_cardinality<decltype(rng0)>::value == unknown);
+		static_assert(ranges::RandomAccessRange<decltype(rng0)>);
+		//  ::models_not<concepts::BoundedView>(aux::copy(rng0));
+		static_assert(ranges::RandomAccessIterator<decltype(rng0.begin())>);
```
CaseyCarter: This is redundant with 66.

test/view/drop_while_view.cpp
```diff
@@ -0,0 +1,87 @@
...
+	//  CHECK_EQUAL(mutable_only, {5,6,7,8,9});
+	//  static_assert(View<decltype(mutable_only)>());
+	//  static_assert(!View<decltype(mutable_only) const>());
+
+	// Check with a mutable predicate
+	{
+		auto rng0 = view::iota(10) | view::drop_while([](int i) mutable { return i < 25; });
+		//  static_assert(range_cardinality<decltype(rng0)>::value == unknown);
+		static_assert(ranges::RandomAccessRange<decltype(rng0)>);
+		//  ::models_not<concepts::BoundedView>(aux::copy(rng0));
```
CaseyCarter: This could be `static_assert(!CommonRange<decltype(rng0)>);`

test/view/take_while_view.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+	std::vector<int> vi{0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+	auto rng1 = vi | view::take_while([](int i) { return i != 50; });
+	static_assert(ranges::RandomAccessRange<decltype(rng1)>);
+	CHECK_EQUAL(rng1, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9});
+
+	/////////////////////////////////////////////////////////////////////////////////////////////////
+	//																															  //
+	// 						DISABLED because the lambda is not equality-preserving.						  //
+	//							Left in the review only to confirm that this is the case.					  //
```
CaseyCarter: +1

test/view/take_while_view.cpp
```diff
@@ -0,0 +1,92 @@
...
+	// int rgi[] = {0,1,2,3,4,5,6,7,8,9};
+	// int cnt = 0;
+	// auto mutable_only = view::take_while(rgi, [cnt](int) mutable { return ++cnt <= 5;});
+	// CHECK_EQUAL(mutable_only, {0,1,2,3,4});
+	// static_assert(ranges::View<decltype(mutable_only)>);
+	// static_assert(!ranges::View<decltype(mutable_only) const>);
+
+	// Check with a mutable predicate
+	{
+		auto rng0 = view::iota(10) | view::take_while([](int i) mutable { return i != 25; });
```
CaseyCarter: Ditto "should actually mutate something."

include/stl2/view/take_while.hpp
CaseyCarter: The easiest fix would be to always require the predicate to be `const`-invocable. We could vastly complicate the view to allow non-`const`-invocable predicates by adding a specialization for them that implements a custom input iterator and caches the predicate evaluation in the view object on first `begin` and iterator increment, but I suspect the added functionality wouldn't be worth the trouble to implement.

cjdb: What is the Standard likeliest to adopt?

test/view/drop_while_view.cpp
cjdb: It's evil, and I like it! 😈

test/view/take_while_view.cpp
cjdb: Doing this means that the lambda isn't `const`-invocable, which causes waves for `const Pred*` in the sentinel. Would making it non-`const` pose any problems for [[res.on.data.races]](http://eel.is/c++draft/res.on.data.races)?

(This isn't a problem for `drop_while_view`, because it doesn't handle its predicate in the same way).

include/stl2/view/take_while.hpp
CaseyCarter: Whatever we can convince a roomful of people is useful in practice. Without some motivating use cases for my whackadoo specialization idea, it will get cut. I'm inclined to always require `const`-invocable and keep it simple.

test/view/take_while_view.cpp
CaseyCarter: If we allow non-`const`-invocable lambdas, the `mutable` lambda test should actually mutate something. If we forbid them, we can't test with a `mutable` lambda because it won't be `const`-invocable whether it actually mutates something or not.

include/stl2/view/drop_while.hpp
```diff
@@ -54,7 +53,21 @@ STL2_OPEN_NAMESPACE {
...
+						// check that te predicate isn't equality-preserving, we can trap
+						// non-equality-preserving invocables on-the-spot by calling them multiple times
+						// and compare the results.
+						// This is only enabled in debug-mode, since it's potentially an expensive check.
+						auto result1 = get()(i);
+						auto result2 = get()(i);
+						STL2_ASSERT(result1 == result2);
+						return result1;
+#else
+						return get()(i);
```
CaseyCarter: I think we can avoid the `#if` and simplify this all down to:
```c++
bool result = get()(i);
STL2_ASSERT(result == get()(i));
return result;
```
`Boolean` requires the implicit conversion to `bool` (and comparison with `bool`) to work, and we don't need to worry that holding a `bool` for a bit and then returning it is less efficient than might be the case for an arbitrary model of `Boolean` returned by a predicate.

include/stl2/view/drop_while.hpp
```diff
@@ -65,7 +78,7 @@ STL2_OPEN_NAMESPACE {
 			R base_;
 		};
 
-		template <Range R, IndirectPredicate<iterator_t<R>> Pred>
+		template <class R, ext::CopyConstructibleObject Pred>
```
CaseyCarter: `ext::CopyConstructibleObject` shouldn't be necessary here: the `drop_while_view` specialization applies that constraint for us. (Well, sort of. It requires `CopyConstructible && std::is_object_v` which is equal to `ext::CopyConstructible` but not equivalent to subsumption in post-[P0717](https://wg21.link/P0717) world. Since subsumption isn't involved here anywhere, the difference is moot.)

include/stl2/view/take_while.hpp
```diff
@@ -67,24 +69,25 @@ STL2_OPEN_NAMESPACE {
 			}
 		};
 
-		template <Range R, IndirectPredicate<iterator_t<R>> Pred>
+		template <class R, ext::CopyConstructibleObject Pred>
```
CaseyCarter: Ditto "let the view impose the constraints so we don't need to repeat them."

test/view/drop_while_view.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+int main()
+{
+	{
+		auto rng0 = view::iota(10) | view::drop_while([](int i) { return i < 25; });
+		//  static_assert(range_cardinality<decltype(rng0)>::value == unknown);
+		static_assert(ranges::RandomAccessRange<decltype(rng0)>);
+		static_assert(!ranges::CommonRange<decltype(rng0)>);
+		static_assert(ranges::RandomAccessIterator<decltype(rng0.begin())>);
```
CaseyCarter: This is still redundant with line 31. (We're not testing the range concepts here, and `RandomAccessRange`s always have `RandomAccessIterator`s.)

test/view/take_while_view.cpp
```diff
@@ -0,0 +1,78 @@
...
+	using namespace ranges::view;
+	using view::ext::take_while;
+} // namespace view
+
+int main()
+{
+	auto rng0 = view::iota(10) | view::take_while([](int i) { return i != 25; });
+	CHECK_EQUAL(rng0, {10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24});
+	static_assert(ranges::View<decltype(rng0)>);
+	// ::models_not<concepts::BoundedView>(aux::copy(rng0));
```
CaseyCarter: You could transform this to `!CommonRange` as well.

test/view/take_while_view.cpp
cjdb: I think this is the only blocker now. I'm hesitant to forbid non-`const`-invocable lambdas, because it eliminates caches, and I feel like that's a good enough reason to go to the effort of allowing them (caches don't preclude equality-preservation, right?).

cjdb: Having said that, I can't think of any motivating uses for a cache, so maybe it belongs on the backlog for when it actually becomes necessary.

CaseyCarter: I've been waffling back and forth on this after spending some time messing with `transform_view` yesterday. There's a tradeoff between "Do we want to make it easy to use correct non-`const`-invocable function objects" and "Do we want to make it hard to use incorrect non-`const`-invocable function objects." I doubt that users understand [res.on.data.races] or the consequences for their function objects that mutate internal state: if we simply accept non-`const`-invocable function objects and assume that the user is defending their internal state against data races, users *will* write bugs. If we only accept `const`-invocable function objects, then motivated users who *really* want mutable internal state will have to jump through hoops to write function objects with `mutable` members and more likely at least understand that they're doing something dangerous, if not necessarily *why* it's dangerous.

Issue #34, opened 2016-07-10
Missing `operator*` in basic_iterator


asutton:
Some concept, somewhere, uses `*declval<const I&>()` with `I` being a `basic_iterator`. Because it doesn't have a `const` overload of `operator*`, the program is ill-formed.

Patch here: https://gist.github.com/asutton/c381a942e02c3f4c8335cfc25d57bd00

CaseyCarter:
`Readable` requires a `const` overload of `operator*`, `Writable` allows `operator*` to mutate. If "some concept, somewhere" thinks that a non-`Readable` iterator must have `operator*() const`, *that* is the bug that needs to be fixed.

I'll respond to the separate email to work out how to repro this.

asutton:
Built using this version of GCC: https://github.com/gcc-mirror/gcc/tree/jason/concepts-rewrite. Include the disable-opt.diff and tsubst.diff patches from the GCC patches list https://gcc.gnu.org/ml/gcc-patches/2016-07/msg00464.html.

The error appears to originate from `counted_iterator` when compiling (at least) the generate.cpp test case. The `noexcept` specifier includes `*declval<const I&>())`, which fails to find a const overload for `basic_iterator`.

In file included from /Users/Sutton/Code/cmcstl2/include/stl2/iterator.hpp:25:0,
                 from /Users/Sutton/Code/cmcstl2/include/stl2/detail/algorithm/generate.hpp:15,
                 from /Users/Sutton/Code/cmcstl2/test/algorithm/generate.cpp:21:
/Users/Sutton/Code/cmcstl2/include/stl2/detail/iterator/counted_iterator.hpp: In instantiation of ‘decltype(auto) std::experimental::ranges::v1::__counted_iterator::cursor<I>::read() const requires (InputIterator<I>)() [with I = std::experimental::ranges::v1::basic_iterator<std::experimental::ranges::v1::detail::back_insert_cursor<std::vector<int> > >]’:
/Users/Sutton/Code/cmcstl2/include/stl2/detail/iterator/basic_iterator.hpp:164:9:   required from ‘class std::experimental::ranges::v1::basic_iterator<std::experimental::ranges::v1::__counted_iterator::cursor<std::experimental::ranges::v1::basic_iterator<std::experimental::ranges::v1::detail::back_insert_cursor<std::vector<int> > > > >’
[...]

PR #181, opened 2018-08-06
add a benchmark for ericniebler/stl2#34

ericniebler:
Thusly proving that @tomaszkam's "itinerary" example can be implemented efficiently in STL2 with very minor changes, leaving the `CommonReference` requirement intact on `StrictTotallyOrderedWith`.

The benchmark, compiled with gcc-8.1 on a MacBook, gives me the following results:

```
2018-08-06 15:03:41
Running benchmark/itinerary
Run on (8 X 2800 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
-----------------------------------------------------------------------
Benchmark                                Time           CPU Iterations
-----------------------------------------------------------------------
ItineraryFixture/STL1/1024/1            30 ns         30 ns   22817059
ItineraryFixture/STL1/4096/1            37 ns         37 ns   18764693
ItineraryFixture/STL1/32768/1           59 ns         59 ns   13328002
ItineraryFixture/STL1/262144/1          79 ns         78 ns    9318052
ItineraryFixture/STL1/1048576/1         97 ns         94 ns    8003842
ItineraryFixture/STL1/1024/2            30 ns         30 ns   21461070
ItineraryFixture/STL1/4096/2            38 ns         38 ns   18467075
ItineraryFixture/STL1/32768/2           51 ns         51 ns   13753807
ItineraryFixture/STL1/262144/2          68 ns         68 ns   10493966
ItineraryFixture/STL1/1048576/2         87 ns         87 ns    7763717
ItineraryFixture/STL2/1024/1            26 ns         26 ns   27363845
ItineraryFixture/STL2/4096/1            31 ns         31 ns   23047544
ItineraryFixture/STL2/32768/1           45 ns         45 ns   15155583
ItineraryFixture/STL2/262144/1          62 ns         62 ns   11136212
ItineraryFixture/STL2/1048576/1         72 ns         71 ns    9794045
ItineraryFixture/STL2/1024/2            24 ns         24 ns   29091272
ItineraryFixture/STL2/4096/2            31 ns         31 ns   22127950
ItineraryFixture/STL2/32768/2           45 ns         45 ns   14772140
ItineraryFixture/STL2/262144/2          66 ns         66 ns   10266037
ItineraryFixture/STL2/1048576/2         76 ns         76 ns    8112086
```

That is, according to the benchmark, STL2 is actually _slightly faster_ than the STL1 code. Go figure.

This relates to ericniebler/stl2#34.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "external/google-benchmark"]
+	path = external/google-benchmark
+	url = https://github.com/google/benchmark.git
+
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,7 @@ install(
     FILES ${PROJECT_BINARY_DIR}/cmcstl2-config.cmake
     DESTINATION lib/cmake/cmcstl2)
 
+add_subdirectory(benchmark)
 add_subdirectory(examples)
 
 enable_testing()
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -1,0 +1,19 @@
+# cmcstl2 - A concept-enabled C++ standard library
+#
+#  Copyright Eric Niebler 2018
+#
+#  Use, modification and distribution is subject to the
+#  Boost Software License, Version 1.0. (See accompanying
+#  file LICENSE_1_0.txt or copy at
+#  http://www.boost.org/LICENSE_1_0.txt)
+#
+# Project home: https://github.com/caseycarter/cmcstl2
+#
+
+find_package(Threads REQUIRED)
+
+include_directories("../external/google-benchmark/include")
+file(GLOB GOOGLE_BENCHMARK_FILES "../external/google-benchmark/src/*.cc")
+
+add_executable(itinerary itinerary.cpp ${GOOGLE_BENCHMARK_FILES})
+target_link_libraries(itinerary stl2 Threads::Threads)
--- a/benchmark/itinerary.cpp
+++ b/benchmark/itinerary.cpp
@@ -1,0 +1,173 @@
+// cmcstl2 - A concept-enabled C++ standard library
+//
+// Copyright Tomasz Kamiński 2016
+// Copyright Eric Niebler 2018
+//
+// Use, modification and distribution is subject to the
+// Boost Software License, Version 1.0. (See accompanying
+// file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//
+// Project home: https://github.com/caseycarter/cmcstl2
+
+#include <vector>
+#include <algorithm>
+#include <random>
+#include <array>
+#include <benchmark/benchmark.h>
+#include <experimental/ranges/range>
+#include <experimental/ranges/algorithm>
+
+namespace std::experimental::ranges
+{
+    template<class Ref>
+    struct any_view
+    {
+        any_view() = default;
+        template<class R>
+          requires !Same<decay_t<R>, any_view> && Range<R> &&
+            ConvertibleTo<iter_reference_t<iterator_t<R>>, Ref>
+        any_view(R&&)
+        {}
+        std::add_pointer_t<Ref> begin() const { return 0; }
+        std::add_pointer_t<Ref> end() const { return 0; }
+    };
+    template <Range R1, Range R2>
+      requires CommonReference<iter_reference_t<iterator_t<R1>>,
+                               iter_reference_t<iterator_t<R2>>>
+    struct common_type<R1, R2>
+    {
+        using type =
+            any_view<common_reference_t<iter_reference_t<iterator_t<R1>>,
+                                        iter_reference_t<iterator_t<R2>>>>;
+    };
+}
+
+namespace ranges = std::experimental::ranges;
+
+struct Date
+{
+    int v;
+};
+
+bool operator==(Date d1, Date d2) { return d1.v == d2.v; }
+bool operator!=(Date t1, Date t2) { return !(t1 == t2); }
+
+bool operator<(Date d1, Date d2) { return d1.v < d2.v; }
+bool operator>(Date t1, Date t2) { return t1 < t2; }
+bool operator>=(Date t1, Date t2) { return !(t1 < t2); }
+bool operator<=(Date t1, Date t2) { return !(t2 < t1); }
+
+struct Leg
+{
+    Date d;
+    Date departure_date() const { return d; }
+};
+
+struct Itinerary
+{
+    std::vector<Leg> l;
+    std::vector<Leg> const& legs() const { return l; }
+};
+
+struct LegDepartureComparator
+{
+    bool operator()(Date d, Leg const& l) const
+    { return d < l.departure_date(); }
+
+    bool operator()(Leg const& l, Date d) const
+    { return  l.departure_date() < d; }
+
+    bool operator()(Leg const& l1, Leg const& l2) const
+    { return l1.departure_date() < l2.departure_date(); }
+};
+
+struct ItineraryDepartureComparator
+{
+   bool operator()(std::vector<Date> const& ds, Itinerary const& i) const
+   { return std::lexicographical_compare(ds.begin(), ds.end(),
+                                         i.legs().begin(), i.legs().end(),
+                                         LegDepartureComparator()); }
+
+   bool operator()(Itinerary const& i, std::vector<Date> const& ds) const
+   { return std::lexicographical_compare(i.legs().begin(), i.legs().end(),
+                                          ds.begin(), ds.end(),
+                                          LegDepartureComparator()); }
+
+   bool operator()(Itinerary const& i1, Itinerary const& i2) const
+   { return std::lexicographical_compare(i1.legs().begin(), i1.legs().end(),
+                                         i2.legs().begin(), i2.legs().end(),
+                                         LegDepartureComparator()); }
+};
+
+class ItineraryFixture : public ::benchmark::Fixture {
+public:
+    void SetUp(const ::benchmark::State& st)
+    {
+        std::size_t const elements = st.range(0);
+        std::size_t const legs = st.range(1);
+
+        std::mt19937 gen;
+        std::uniform_int_distribution<int> dist(0, 30);
+
+        dates.resize(legs);
+        for (Date& d : dates)
+          d.v = dist(gen);
+
+        itineraries.resize(elements);
+        for (Itinerary& itin : itineraries)
+        {
+           itin.l.resize(legs);
+           for (Leg& leg : itin.l)
+              leg.d.v = dist(gen);
+        }
+        std::sort(itineraries.begin(), itineraries.end(),
+                  ItineraryDepartureComparator());
+    }
+
+    void TearDown(const ::benchmark::State&)
+    {
+        itineraries.clear();
+        dates.clear();
+    }
+
+   std::vector<Itinerary> itineraries;
+   std::vector<Date> dates;
+};
+
+BENCHMARK_DEFINE_F(ItineraryFixture, STL1)(benchmark::State& state)
+{
+    while (state.KeepRunning())
+    {
+        benchmark::DoNotOptimize(
+            std::equal_range(itineraries.begin(), itineraries.end(),
+                             dates, ItineraryDepartureComparator()));
+    }
+}
+BENCHMARK_REGISTER_F(ItineraryFixture, STL1)
+  -> Ranges({{1<<10, 1<<20}, {1,2}});
+
+inline constexpr auto toDate = [](Leg const& l)
+{
+    return l.departure_date();
+};
+inline constexpr auto toDates = [](Itinerary const& i)
+{
+    return i.legs() |
+        ranges::view::transform(toDate);
+};
+
+BENCHMARK_DEFINE_F(ItineraryFixture, STL2)(benchmark::State& state)
+{
+   while (state.KeepRunning())
+   {
+        benchmark::DoNotOptimize(
+            ranges::equal_range(
+                itineraries,
+                ranges::view::all(dates),
+                [](auto const& r1, auto const& r2) { return ranges::lexicographical_compare(r1, r2); },
+                toDates));
+    }
+}
+BENCHMARK_REGISTER_F(ItineraryFixture, STL2)
+  -> Ranges({{1<<10, 1<<20}, {1,2}});
--- a/include/stl2/detail/algorithm/lexicographical_compare.hpp
+++ b/include/stl2/detail/algorithm/lexicographical_compare.hpp
@@ -21,41 +21,44 @@
 // lexicographical_compare [alg.lex.comparison]
 //
 STL2_OPEN_NAMESPACE {
-	template<InputIterator I1, Sentinel<I1> S1, InputIterator I2, Sentinel<I2> S2,
-		class Comp = less<>, class Proj1 = identity, class Proj2 = identity>
-	requires
-		IndirectStrictWeakOrder<Comp,
-			projected<I1, Proj1>, projected<I2, Proj2>>
-	bool lexicographical_compare(I1 first1, S1 last1, I2 first2, S2 last2,
-		Comp comp = Comp{}, Proj1 proj1 = Proj1{}, Proj2 proj2 = Proj2{})
+	inline constexpr struct lexicographical_compare_fn
 	{
-		for (; first1 != last1 && first2 != last2; ++first1, ++first2) {
-			if (__stl2::invoke(comp, __stl2::invoke(proj1, *first1), __stl2::invoke(proj2, *first2))) {
-				return true;
+		template <InputIterator I1, Sentinel<I1> S1, InputIterator I2, Sentinel<I2> S2,
+			class Comp = less<>, class Proj1 = identity, class Proj2 = identity>
+		requires
+			IndirectStrictWeakOrder<Comp,
+				projected<I1, Proj1>, projected<I2, Proj2>>
+		bool operator()(I1 first1, S1 last1, I2 first2, S2 last2,
+			Comp comp = Comp{}, Proj1 proj1 = Proj1{}, Proj2 proj2 = Proj2{}) const
+		{
+			for (; first1 != last1 && first2 != last2; ++first1, ++first2) {
+				if (__stl2::invoke(comp, __stl2::invoke(proj1, *first1), __stl2::invoke(proj2, *first2))) {
+					return true;
+				}
+				if (__stl2::invoke(comp, __stl2::invoke(proj2, *first2), __stl2::invoke(proj1, *first1))) {
+					return false;
+				}
 			}
-			if (__stl2::invoke(comp, __stl2::invoke(proj2, *first2), __stl2::invoke(proj1, *first1))) {
-				return false;
-			}
+			return first1 == last1 && first2 != last2;
 		}
-		return first1 == last1 && first2 != last2;
-	}
 
-	template<InputRange Rng1, InputRange Rng2, class Comp = less<>,
-		class Proj1 = identity, class Proj2 = identity>
-	requires
-		IndirectStrictWeakOrder<Comp,
-			projected<iterator_t<Rng1>, Proj1>,
-			projected<iterator_t<Rng2>, Proj2>>
-	bool lexicographical_compare(Rng1&& rng1, Rng2&& rng2,
-		Comp comp = Comp{}, Proj1 proj1 = Proj1{}, Proj2 proj2 = Proj2{})
-	{
-		return __stl2::lexicographical_compare(
-			__stl2::begin(rng1), __stl2::end(rng1),
-			__stl2::begin(rng2), __stl2::end(rng2),
-			std::ref(comp),
-			std::ref(proj1),
-			std::ref(proj2));
-	}
+		template <InputRange Rng1, InputRange Rng2, class Comp = less<>,
+			class Proj1 = identity, class Proj2 = identity>
+		requires
+			IndirectStrictWeakOrder<Comp,
+				projected<iterator_t<Rng1>, Proj1>,
+				projected<iterator_t<Rng2>, Proj2>>
+		bool operator()(Rng1&& rng1, Rng2&& rng2,
+			Comp comp = Comp{}, Proj1 proj1 = Proj1{}, Proj2 proj2 = Proj2{}) const
+		{
+			return (*this)(
+				__stl2::begin(rng1), __stl2::end(rng1),
+				__stl2::begin(rng2), __stl2::end(rng2),
+				std::ref(comp),
+				std::ref(proj1),
+				std::ref(proj2));
+		}
+	} const lexicographical_compare {};
 } STL2_CLOSE_NAMESPACE
 
 #endif


CaseyCarter:
> Also, this solution provides strong motivation for providing relational operators for ranges, which default to `ranges::equal` (for `==` and `!=`) and `ranges::lexicographical_compare` (for `<`, `>`, `<=`, and `>=`).

Egads no. WG21 *just* got done arguing that `span` shouldn't have an equality operator that compares the elements it denotes because they are subject to change by operations that don't modify the `span`. The element values cannot be salient to the value of a range that doesn't own those elements.

ericniebler:
Oops, you're right about that. :-P

ericniebler:
I can implement the same benchmark in range-v3 and show the same effect on any compiler (capable of compiling range-v3, that is). The only reason you were seeing any slow-down at all was because your STL1 code wasn't using pointers-to-members and your STL2 code was. That is, you were comparing apples to oranges. (Compilers do a poor job of optimizing pointers-to-members, possibly because they are not that common in end-user code.)

As for that old blog post, there are some views that are difficult for some compilers to optimize; `view::join` for instance, which is used to implement `view::for_each`. On the other hand, `view::transform` is pretty easy for compilers to optimize. I wouldn't use that post to conclude that ranges are inefficient or difficult to optimize in general. And it's certainly not the case that the `CommonReference` requirement on cross-type concepts has anything to do with clang's inability to optimize the Pythagorean triples example.

tomaszkam:
Few things regarding the request:
1) The code has UB, as a comparison of any_view does not match the comparison of the algorithms, i.e. the semantics requirements of StrictWeakOrdering are not satisfied.. This is important because the `any_view` is currently not proposed, so maybe not available for the user of C++20.
2) To make code shorter (compared to STL1) you are using fact that algorithms are functors (not functions), which again is not true for C++20.
So, please provide a solution that actually works with current working-draft for C++20, or it is proposed to be there. Any other option may (and probably won't be available) for programmers.

tomaszkam:
The fact is I am unable to reproduce your result on my system. I am using g++-7:
```Using built-in specs.
COLLECT_GCC=g++-7
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/7/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 7.3.0-16ubuntu3' --with-bugurl=file:///usr/share/doc/gcc-7/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --with-as=/usr/bin/x86_64-linux-gnu-as --with-ld=/usr/bin/x86_64-linux-gnu-ld --program-suffix=-7 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 7.3.0 (Ubuntu 7.3.0-16ubuntu3)
```

Then the program is compiled as follows:
`g++-7 -O2 -std=c++17 -fconcepts itinerary.cpp -Icmcstl2/include/ -lbenchmark -lpthread`
I.e. optimization level that the projects I am working on are usually build with,

And I get the following result on my PC with Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz:
```2018-08-07 20:38:26
Running ./a.out
Run on (8 X 4200 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
-----------------------------------------------------------------------
Benchmark                                Time           CPU Iterations
-----------------------------------------------------------------------
ItineraryFixture/STL1/1024/1            26 ns         26 ns   27150714
ItineraryFixture/STL1/4096/1            33 ns         33 ns   21486153
ItineraryFixture/STL1/32768/1           52 ns         52 ns   14354169
ItineraryFixture/STL1/262144/1          65 ns         65 ns   10948499
ItineraryFixture/STL1/1048576/1         86 ns         86 ns    8321704
ItineraryFixture/STL1/1024/2            24 ns         24 ns   29284085
ItineraryFixture/STL1/4096/2            31 ns         31 ns   22932089
ItineraryFixture/STL1/32768/2           47 ns         47 ns   16559947
ItineraryFixture/STL1/262144/2          59 ns         59 ns   12459084
ItineraryFixture/STL1/1048576/2         79 ns         79 ns    9176410
ItineraryFixture/STL2/1024/1            40 ns         40 ns   17388324
ItineraryFixture/STL2/4096/1            47 ns         47 ns   14701711
ItineraryFixture/STL2/32768/1           64 ns         64 ns   10895889
ItineraryFixture/STL2/262144/1          77 ns         77 ns    9109058
ItineraryFixture/STL2/1048576/1         98 ns         98 ns    7005623
ItineraryFixture/STL2/1024/2            36 ns         36 ns   19695671
ItineraryFixture/STL2/4096/2            43 ns         43 ns   16237604
ItineraryFixture/STL2/32768/2           58 ns         58 ns   11944594
ItineraryFixture/STL2/262144/2          78 ns         78 ns    9373197
ItineraryFixture/STL2/1048576/2         98 ns         98 ns    7356964
```

ericniebler:
> The code has UB, as a comparison of any_view does not match the comparison of the algorithms

I took a shortcut. Of course a full solution would completely flesh out `any_view`, and there is nothing hard about its implementation.

> you are using fact that algorithms are functors (not functions), which again is not true for C++20

They are (disguised) function objects, so my code will work as written in C++20. (We're a little circumspect in the current draft about whether the algorithms in `std::ranges` are function objects or not, but in truth there is no other way to implement them.)

And here is my result for gcc-7.3, again showing no difference. I use `-O3 -DNDEBUG`, so maybe that's what you're seeing?

```
2018-08-07 12:05:11
Running ./benchmark/itinerary
Run on (8 X 2800 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 262K (x4)
  L3 Unified 6291K (x1)
-----------------------------------------------------------------------
Benchmark                                Time           CPU Iterations
-----------------------------------------------------------------------
ItineraryFixture/STL1/1024/1            33 ns         33 ns   20874710
ItineraryFixture/STL1/4096/1            40 ns         40 ns   16970890
ItineraryFixture/STL1/32768/1           63 ns         63 ns   12080838
ItineraryFixture/STL1/262144/1          80 ns         80 ns    8872440
ItineraryFixture/STL1/1048576/1         94 ns         94 ns    7655126
ItineraryFixture/STL1/1024/2            31 ns         31 ns   22644122
ItineraryFixture/STL1/4096/2            38 ns         38 ns   18323168
ItineraryFixture/STL1/32768/2           54 ns         54 ns   12208096
ItineraryFixture/STL1/262144/2          72 ns         72 ns    9740757
ItineraryFixture/STL1/1048576/2         90 ns         90 ns    7595404
ItineraryFixture/STL2/1024/1            26 ns         26 ns   26716741
ItineraryFixture/STL2/4096/1            32 ns         32 ns   21591344
ItineraryFixture/STL2/32768/1           50 ns         50 ns   13999720
ItineraryFixture/STL2/262144/1          65 ns         65 ns   10741963
ItineraryFixture/STL2/1048576/1         74 ns         74 ns    9471363
ItineraryFixture/STL2/1024/2            23 ns         23 ns   29850619
ItineraryFixture/STL2/4096/2            31 ns         31 ns   23021999
ItineraryFixture/STL2/32768/2           46 ns         46 ns   15243637
ItineraryFixture/STL2/262144/2          65 ns         65 ns   10831386
ItineraryFixture/STL2/1048576/2         79 ns         79 ns    8998355
```

Again, the STL2 code is shorter, simpler, more readable, and faster, _even with the `CommonReference` requirement_. The STL1 code defines semantically meaningless relations over apples and oranges. I see no reason for it.

CaseyCarter:
I suggest you both test with the implementation on the `benchmark_fixes` branch, which eliminates variance between the library implementations by using the cmcstl2 algorithms for both the "STL1" and "STL2" benchmarks. (Obviously the requirements for the "STL1" code have been largely removed.)

CaseyCarter:
@tomaszkam, I also suggest you compile `-DNDEBUG` to disable the debug checking in cmcstl2.

tomaszkam:
Putting -NDEBUG with -O2 does not help:
```> g++-7 -O2 -DNDEBUG  -std=c++17 -fconcepts itinerary.cpp -Icmcstl2/include/ -lbenchmark -lpthread
> ./a.out 
2018-08-07 21:37:03
Running ./a.out
Run on (8 X 4200 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
-----------------------------------------------------------------------
Benchmark                                Time           CPU Iterations
-----------------------------------------------------------------------
ItineraryFixture/STL1/1024/1            26 ns         26 ns   27149589
ItineraryFixture/STL1/4096/1            33 ns         33 ns   21467442
ItineraryFixture/STL1/32768/1           50 ns         50 ns   15329865
ItineraryFixture/STL1/262144/1          65 ns         65 ns   10858852
ItineraryFixture/STL1/1048576/1         82 ns         82 ns    8905017
ItineraryFixture/STL1/1024/2            24 ns         24 ns   29270811
ItineraryFixture/STL1/4096/2            31 ns         31 ns   22936784
ItineraryFixture/STL1/32768/2           46 ns         46 ns   16075759
ItineraryFixture/STL1/262144/2          59 ns         59 ns   12132677
ItineraryFixture/STL1/1048576/2         78 ns         78 ns    9288086
ItineraryFixture/STL2/1024/1            37 ns         37 ns   19111440
ItineraryFixture/STL2/4096/1            46 ns         46 ns   14923145
ItineraryFixture/STL2/32768/1           64 ns         64 ns   11011531
ItineraryFixture/STL2/262144/1          79 ns         79 ns    8617140
ItineraryFixture/STL2/1048576/1        107 ns        107 ns    6865674
ItineraryFixture/STL2/1024/2            31 ns         31 ns   22421557
ItineraryFixture/STL2/4096/2            42 ns         42 ns   16700593
ItineraryFixture/STL2/32768/2           56 ns         56 ns   12402371
ItineraryFixture/STL2/262144/2          74 ns         74 ns    9649730
```

Bumping up optimization level to O3, at least give me consistent results with one from Eric:
```> g++-7 -O3 -DNDEBUG  -std=c++17 -fconcepts itinerary.cpp -Icmcstl2/include/ -lbenchmark -lpthread
> ./a.out 
2018-08-07 21:39:21
Running ./a.out
Run on (8 X 4200 MHz CPU s)
CPU Caches:
  L1 Data 32K (x4)
  L1 Instruction 32K (x4)
  L2 Unified 256K (x4)
  L3 Unified 8192K (x1)
-----------------------------------------------------------------------
Benchmark                                Time           CPU Iterations
-----------------------------------------------------------------------
ItineraryFixture/STL1/1024/1            26 ns         26 ns   26607868
ItineraryFixture/STL1/4096/1            33 ns         33 ns   21068874
ItineraryFixture/STL1/32768/1           49 ns         49 ns   14726723
ItineraryFixture/STL1/262144/1          65 ns         65 ns   10625095
ItineraryFixture/STL1/1048576/1         78 ns         78 ns    8931364
ItineraryFixture/STL1/1024/2            25 ns         25 ns   28320784
ItineraryFixture/STL1/4096/2            31 ns         31 ns   22362341
ItineraryFixture/STL1/32768/2           47 ns         47 ns   16273846
ItineraryFixture/STL1/262144/2          59 ns         59 ns   12040877
ItineraryFixture/STL1/1048576/2         80 ns         80 ns    9351142
ItineraryFixture/STL2/1024/1            22 ns         22 ns   31693527
ItineraryFixture/STL2/4096/1            27 ns         27 ns   25750684
ItineraryFixture/STL2/32768/1           42 ns         42 ns   17059459
ItineraryFixture/STL2/262144/1          53 ns         53 ns   12459202
ItineraryFixture/STL2/1048576/1         71 ns         70 ns   10075356
ItineraryFixture/STL2/1024/2            19 ns         19 ns   36026087
ItineraryFixture/STL2/4096/2            24 ns         24 ns   29528166
ItineraryFixture/STL2/32768/2           40 ns         40 ns   18340397
ItineraryFixture/STL2/262144/2          57 ns         57 ns   13138066
ItineraryFixture/STL2/1048576/2         70 ns         70 ns   10785384
```
However, change in optimization level may not be an option for some of the projects.

CaseyCarter:
> They are (disguised) function objects, so my code will work as written in C++20. (We're a little circumspect in the current draft about whether the algorithms in `std::ranges` are function objects or not, but in truth there is no other way to implement them.)

An implementation is in no way required to implement them as function objects: it's only required that they are not findable by ADL and inhibit ADL when found via unqualified name lookup. There's no other *pure library* implementation, but a front end developer could throw together a custom attribute to provide those properties for function templates in a day. Your code may or may not work as written in C++20.

CaseyCarter:
> Putting -NDEBUG with -O2 does not help:

Is that with `itinerary.cpp` from #182? I can't explain the differences we're seeing from the same command line and almost precisely the same compiler.

tomaszkam:
> Is that with itinerary.cpp from #182? I can't explain the differences we're seeing from the same command line and almost precisely the same compiler.

I changed the:
```benchmark::DoNotOptimize(
     ranges::equal_range(
         itineraries,
         dates,
         ranges::lexicographical_compare.
        toDates));
```
To something that is guaranteed to work with standard, i.e.:
```benchmark::DoNotOptimize(
     ranges::equal_range(
         itineraries,
         dates,
         [](auto const& r1, auto const& r2) { return ranges::lexicographical_compare(r1, r2); },
        toDates));
```

ericniebler:
@CaseyCarter wrote:
> An implementation is in no way required to implement them as function objects... Your code may or may not work as written in C++20.

Bah, humbug! Being able to pass functions to functions is fundamental and very useful. It galls me that implementors _will_ write them as objects (mark my words &lt;shakes cane&gt;), but we're not allowed to treat them as such.

For the record, I see no performance impact after lifting `ranges::lexicographical_compare` into a trivial lambda.

@tomaszkam wrote:
> the difference is caused by the usage of -O2 vs -O3, however, the apps I have contact with consistently use O2, for them, we are seeing performance drop caused by compiler not optimizing the code - exactly the point I made in paper.

If the argument you are bringing to San Diego is that you should be free to write worse code because it optimizes a little better at lower optimization settings, I can tell you how that argument will go over.

Don't get me wrong. You have made some good points in your paper; in particular the unfortunate integer promotion rules and their interaction with the semantic constraints of `Relation` are a legitimate problem. Your argument about performance is not helping your case, though. You should either find a better example that substantiates your claim about performance or drop that argument. Your paper will be better for it.

CaseyCarter:
> I changed ... to something that is guaranteed to work with standard

I've made a similar change to #182 (I stored the lambda in a named variable outside the benchmark loop). I still have results that differ by at most 5ns at both `-O2` and `-O1` (`-O0` is a very different story: the STL2 benchmarks take ~twice the time).

CaseyCarter:
> Being able to pass functions to functions is fundamental and very useful.

Being able to pass function templates and overload sets to functions would be useful as well. I consider it a defect in C++'s support for the functional paradigm that we can't do this today.

ericniebler:
@tomaszkam One other suggestion for your paper: rather that simply dropping the `CommonReference` syntactic requirements, you could make it a purely semantic requirement instead. Appeal to the existence of a common type that is outside the type system. So long as such a type can be implemented in theory, it need not exist in practice. The relaxation of `Common` in [N3351](http://wg21.link/N3351) hints at that much in D.2, and @sean-parent and @asutton 
have both said as much explicitly. Without something like that, I cannot support your paper (and still may argue against it).

Given such a purely semantic common type constraint on `Relation`, your STL1 code is still questionable.

tomaszkam:
> @tomaszkam One other suggestion for your paper: rather that simply dropping the CommonReference syntactic requirements, you could make it a purely semantic requirement instead. Appeal to the existence of a common type that is outside the type system. So long as such a type can be implemented in theory, it need not exist in practice. The relaxation of Common in N3351 hints at that much in D.2, and @sean-parent and @asutton have both said as much explicitly. Without something like that, I cannot support your paper (and still may argue against it).

But please provide a justification for this requirement in case of the weak orderings, where the object are merely equivalent. The underlying mathematical model allows building relation over different types (the domain is just union of the objects of a different type), the definition placed in the books you are basing your work (Elements of Programming) does also not mention such requirement. From that perspective, this is only the product of your proposal.

The is major difference between weak and strong orderings, as one of the allows an equivalent object (and things are equivalent if they from equivalence relation - there are 3 axioms required for that and nothing more), and other make objects equal (in the mathematical sense same). For the later I agree, that we should be able to point of common_type, as they are meant to represent same underlying entity.

tomaszkam:
I am willing to assume that actually the difference on the result for the test example, is fault on my side. I am also planning to check it on AMD low-end laptop, to see if it may be architecture related.

However, the example was a simplification of the actual use case we have - i.e. comparing operating flights of itineraries. I have prepared an updated example, that compares departures on actual flights (there should be more data to be compared), but want to built it iteratively.

However, I have miserably failed to build STL2 equivalent for that code. My attempt may be found at:
[here](https://pastebin.com/BQay3Dma) - I get compilation errors from view::join.

tomaszkam:
The error I am getting looks like:
```
In file included from cmcstl2/include/stl2/view/all.hpp:21:0,
                 from cmcstl2/include/stl2/ranges.hpp:22,
                 from cmcstl2/include/experimental/ranges/range:12,
                 from itinerary_flights.cpp:18:
cmcstl2/include/stl2/view/ref.hpp: In instantiation of ‘constexpr auto std::experimental::ranges::v1::ext::ref_view<Rng>::size() const requires  SizedRange<Rng> [with Rng = std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_vie
w<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> > >]’:
cmcstl2/include/stl2/detail/range/access.hpp:398:14:   required by substitution of ‘template<class _Tp, class _Arg, class> static std::true_type std::__do_is_direct_constructible_impl::__test(int) [with _Tp = std::experimental::ranges::v1::ext::transform_view<std::exp
erimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> >; _Arg = std::experimental::ranges::v1::ext::ref_view<std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::e
xt::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> > > >; <template-parameter-1-3> = <missing>]’
/usr/include/c++/7/type_traits:984:41:   required from ‘struct std::__is_direct_constructible_impl<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> >, std::experimental::rang
es::v1::ext::ref_view<std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> > > > >’
/usr/include/c++/7/type_traits:143:12:   required from ‘struct std::__and_<std::is_destructible<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg, std::allocator<Leg> > >, <lambda(const Leg&)> > >, st
d::__is_direct_constructible_impl<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg, std::allocator<Leg> > >, <lambda(const Leg&)> >, std::experimental::ranges::v1::ext::ref_view<std::experimental::ra
nges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg, std::allocator<Leg> > >, <lambda(const Leg&)> > > > > >’
/usr/include/c++/7/type_traits:988:12:   required from ‘struct std::__is_direct_constructible_new_safe<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> >, std::experimental::
ranges::v1::ext::ref_view<std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> > > > >’
/usr/include/c++/7/type_traits:1070:12:   required from ‘struct std::__is_direct_constructible_new<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> >, std::experimental::rang
es::v1::ext::ref_view<std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> > > > >’
/usr/include/c++/7/type_traits:1078:12:   [ skipping 14 instantiation contexts, use -ftemplate-backtrace-limit=0 to disable ]
cmcstl2/include/stl2/type_traits.hpp:166:9:   required from ‘struct std::experimental::ranges::v1::__common_reference2<std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const s
td::vector<Leg> >, <lambda(const Leg&)> > >, std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg> >, <lambda(const Leg&)> > >&>’
cmcstl2/include/stl2/type_traits.hpp:175:9:   required from ‘struct std::experimental::ranges::v1::common_reference<std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std:
:vector<Leg, std::allocator<Leg> > >, <lambda(const Leg&)> > >, std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg, std::allocator<Leg> > >, <lambda(const 
Leg&)> > >&>’
cmcstl2/include/stl2/meta/meta.hpp:137:36:   required by substitution of ‘template<class T>  requires  Trait<T> using _t = typename T::type [with T = std::experimental::ranges::v1::common_reference<std::experimental::ranges::v1::ext::join_view<std::experimental::range
s::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<const std::vector<Leg, std::allocator<Leg> > >, <lambda(const Leg&)> > >, std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges
::v1::ext::ref_view<const std::vector<Leg, std::allocator<Leg> > >, <lambda(const Leg&)> > >&>]’
cmcstl2/include/stl2/type_traits.hpp:146:62:   required by substitution of ‘template<class ... Ts> using common_reference_t = meta::v1::_t<std::experimental::ranges::v1::common_reference<Ts ...> > [with Ts = {std::experimental::ranges::v1::indirect_result_t<<lambda(It
inerary&)>&, __gnu_cxx::__normal_iterator<Itinerary*, std::vector<Itinerary, std::allocator<Itinerary> > > >, std::experimental::ranges::v1::readable_traits<std::experimental::ranges::v1::projected<__gnu_cxx::__normal_iterator<Itinerary*, std::vector<Itinerary> >, <la
mbda(Itinerary&)> > >::type&}]’
cmcstl2/include/stl2/detail/iterator/concepts.hpp:179:60:   required by substitution of ‘template<class I>  requires  Readable<I> using iter_common_reference_t = std::experimental::ranges::v1::common_reference_t<decltype (* declval<R&>()), typename std::experimental::
ranges::v1::readable_traits<T>::type&> [with I = std::experimental::ranges::v1::projected<__gnu_cxx::__normal_iterator<Itinerary*, std::vector<Itinerary> >, <lambda(Itinerary&)> >]’
cmcstl2/include/stl2/detail/concepts/callable.hpp:141:3:   required from here
cmcstl2/include/stl2/view/ref.hpp:57:34: error: no match for call to ‘(const std::experimental::ranges::v1::__size::__fn) (std::experimental::ranges::v1::ext::join_view<std::experimental::ranges::v1::ext::transform_view<std::experimental::ranges::v1::ext::ref_view<con
st std::vector<Leg> >, <lambda(const Leg&)> > >&)’
    noexcept(noexcept(__stl2::size(*rng_)))
                      ~~~~~~~~~~~~^~~~~~~
```

tomaszkam:
The difference from the previous example is that now instead of transforming `Itinerary` into `vector<Date>`, I notionally transform it to `std::vector<std::vector<Dates>>` and then flatten it.

The code to construct view looks like follows:
```inline constexpr auto toDate = [](const Flight& f)
{
    return f.departure_date();
};

inline constexpr auto toFlightsDates = [](const Leg& l)
{
    return l.flights() | ranges::view::transform(toDate);
};

inline constexpr auto toDates = [](Itinerary& i)
{
    return ranges::view::join(i.legs() | ranges::view::transform(toFlightsDates));
};
````
Now, I am able to construct such view and iterate over it, but the invocation of `lexicographical_compare` fails, in `iter_common_reference` instatation (this is not intended).

PR #182, opened 2018-08-07
Cleanup CMake horrors

CaseyCarter:


benchmark/itinerary.cpp
```diff
@@ -18,21 +18,173 @@
 #include <experimental/ranges/range>
 #include <experimental/ranges/algorithm>
 
+namespace ranges = std::experimental::ranges;
+
+namespace not_ranges
```
CaseyCarter: This code *will* diverge from what's in the algorithm headers, and render the benchmark useless. The fully-constrained range version of `equal_range` needs to be in this file as well to guarantee that the STL1 and STL2 benchmarks use identical implementations.

CaseyCarter: Fixed in 8d483fb556b48ba665fef3dc030224f79cb82f2c.

Issue #189, opened 2018-08-26
nth_element uses goto


cjdb:
This means it can't be `constexpr`.

CaseyCarter:
Fixed by 361f3560ceca9f073051b50608f4a61d5e6e3861.

Issue #188, opened 2018-08-22
Whither deprecated algorithm overloads?


cjdb:
What should come of the deprecated algorithm overloads, which aren't mentioned in P0896?

Issue #70, opened 2017-01-30
Re-tighten constraints on algorithms


CaseyCarter:
...to *not* require the relaxation specified in [algorithm.general]/12:

> Despite that the algorithm declarations nominally accept parameters by value, it is unspecified when and if the argument expressions are used to initialize the actual parameters except that any such initialization shall be sequenced before (ISO/IEC 14882:2014 §1.9) the algorithm returns.

which will be removed by the PR for ericniebler/stl2#261. Already done for the algorithms with overloads in Annex A.2 (`equal`, `is_permutation`, `mismatch`, `swap_ranges`, `transform`) by b160823ad8d56b488687c0135ea4aa43e3cedcba.

CaseyCarter:
I believe I picked off the few remaining instances of this today.

Issue #41, opened 2016-10-26
Sweep the codebase for components out-of-spec to D0459R1


CaseyCarter:
Anything not integrated by now is likely not important enough to be integrated.

ericniebler:
Yes please. I'm particular, I'm blocked by the not-too-spec way you define `value_type`.

CaseyCarter:
> I'm particular, I'm blocked by the not-too-spec way you define value_type.

Fixed in bb1655783d9573f880d4b62f2cd302bef4ae2d81.

cjdb:
Should functions like `find_if_not` retain their `__f` wrapper?

```cpp
template <class I, class S, class Pred, class Proj = identity> // need to swap Pred and Proj
requires
	InputIterator<__f<I>>() &&
	Sentinel<__f<S>, __f<I>>() &&
	IndirectPredicate<Pred, projected<__f<I>, Proj>>()
__f<I> find_if_not(I&& first, S&& last, Pred pred, Proj proj = Proj{})
{
	return __stl2::find_if(__stl2::forward<I>(first), __stl2::forward<S>(last),
		__stl2::not_fn(__stl2::ref(pred)), __stl2::ref(proj));
}
```

`find_if_not` in N4622 doesn't mention `decay_t`, but I can't project what will happen if it's removed.

CaseyCarter:
It *should* be safe to unwrap those and take the function arguments by value.  Any resulting ambiguities are extensions that I need to move to namespace `ext`.

PR #190, opened 2018-08-26
Converging toward P0896R3 (algorithms and special memory algorithms)

cjdb:
This commit moves cmcstl2 closer toward what will be in P0896R3 for San Diego.

## Changes

* Return types more closely resemble what's in R2. I created exposition-only base structs instead of creating aliases to see how well that would work with the library as a whole.
* Removed a lot of dependency on `tagged.hpp` in the wake of named return types.
* Most algorithms are now implemented as functors, and are not lookupable by ADL, as required by [specialized.algorithms]/3.1 and [algorithms.requirements]/2.
  * Many details now cannot be looked up at all: they're private static members.
  * This potentially eliminates the need for `STL2_WORKAROUND_GCC_79591`, as this bug doesn't seem to surface when a `using` declaration is made in a class definition.
* Algorithm interfaces resemble what is present in ericniebler/stl2@19b9f15f3206d0a5cda9611e3f4345b8f3982d57.
  * Optimisations have been made private static members that are delegated to by the public interface. The intention is to significantly reduce the noise given by diagnostics for people unaware that algorithms can be optimised for certain concepts.
* Some, but not all, of `Rng` has been renamed to `R`.
* `safe_iterator_t` now disallows rvalues: as @CaseyCarter predicted on Slack, this is problematic when passing range adapters into algorithms.

## What's missing?

* I haven't looked at `shuffle` yet, but I imagine it'll follow the same formula. I can add it to this branch, or I can add it to a different one.

## Things that must happen before merging with master:

* The test case for `find_end` is uncommitted (and incomplete) on my work computer, so I've disabled compiling that test to get this PR up early. It'll need to be committed before merging, as there are very likely issues with `find_end` that I haven't had the opportunity to pick up on yet.
* I'd like to rebase and squash this. The commits are useless fluff, mostly so I could change between home and work computers.

## Issues this might close

* #17 constexpr extensions
* #41 Sweep the codebase for components out-of-spec to the Ranges TS WP
* #70 Re-tighten constraints on algorithms
* #188 Whither deprecated algorithm overloads? 

## Issues this pull request opens

* #189 `nth_element` uses `goto`.


cjdb:
Conflicts with master way too much, and it'll be easier to redo this in piecemeal chunks anyway.

PR #192, opened 2018-09-15
Change ebo_box<T>::item_ to be default-initialised

cjdb:
When ebo_box<T> holds a fundamental type, item_'s
value-initialisation prevents default-initialised
objects (presumably inheriting from ebo_box) from
being used in a constexpr context.

This commit default-initialises item_ so that it
is more likely to be usable in constexpr contexts.

### DIFF
--- a/include/stl2/detail/ebo_box.hpp
+++ b/include/stl2/detail/ebo_box.hpp
@@ -54,7 +54,7 @@ STL2_OPEN_NAMESPACE {
 			constexpr const T&& get() const&& noexcept { return std::move(item_); }
 
 		private:
-			T item_;
+			T item_{};
 		};
 
 		template <ext::DestructibleObject T, class Tag>


CaseyCarter:
`ebo_box` uses the "pass-through default init" idiom (TM) to be fully general: it has no user-provided default constructor, and does not initialize `item_`. Consequently, a default-initialized `ebo_box` contains a default-initialized `item_` - allowing to avoid the cost of value-init when desired - and a value-initialized `ebo_box` contains a value-initialized `item_`.

If you want a default-initialized object of a type `T` that inherits from `ebo_box<U>` to be usable in a constant expression, then `T` must value-initialize its `ebo_box<U>` base.

If you want a value-initialized object of a type `T` that inherits from `ebo_box<U>` to be usable in a constant expression, then `T` must either value-initialize its `ebo_box<U>` base or use the pass-through default init idiom itself.

See https://godbolt.org/z/L5od64.

PR #194, opened 2018-09-19
subrange: merge conflicting deduction guides

CaseyCarter:


ericniebler:
Why are these conflicting? Compiler bug?

CaseyCarter:
[temp.deduct.guide]/3 says it's ill-formed to have two deduction guides with identical parameter-lists. I made this change in the proposal so we about making the standard library ill-formed until cwg get this straightened out.

PR #195, opened 2018-10-07
elements: view to extracts elements of tuple-like objects

cor3ntin:
sequence_of_tuple | ranges::ext::view::elements<42>
static_array | ranges::ext::view::elements<0>

map | ranges::ext::view::keys;
map | ranges::ext::view::values;


cor3ntin:
Hi @CaseyCarter, could tell me what you think about this?
The goal was to take a bit of work out of @cjdb for his range paper.


I currently have a weird issue if I try a static_assert(View<Rng>), any idea what I could have done wrong?

```

**Comment by cjdb** [comment]
This is what I've been using in my `zip_with` branch. Is there any reason we can't go with this instead? (I don't _think_ it breaks any equality preservation).

```cpp
namespace view {
	template<int X>
	requires (X == 0) || (X == 1)
	struct __get_fn {
		template<class P>
		requires _PairLike<remove_reference_t<P>> // from subrange.hpp
		constexpr decltype(auto) operator()(P&& p) const
		{
			using T = decltype(std::get<X>(std::forward<P>(p)));
			return std::forward<T>(std::get<X>(std::forward<P>(p)));
		}
	};

	inline constexpr auto keys = view::transform(__get_fn<0>{});
	inline constexpr auto values = view::transform(__get_fn<1>{});
} // namespace view
```

PR #199, opened 2018-10-31
Kill ADL for next, prev, and distance

cjdb:
Same deal as before + a few changes to `Incrementable` so it matches P0896.

I can't do `advance` yet, as `counted_iterator` provides overloads of `advance`. My goal (after San Diego) is to turn `counted_iterator` into `__counting_cursor` and then make a `basic_iterator` out of it.

Getting `distance` to be a function object now would be nice for getting `zip_with` to work.

### DIFF
--- a/include/stl2/detail/iterator/increment.hpp
+++ b/include/stl2/detail/iterator/increment.hpp
@@ -98,7 +98,7 @@ STL2_OPEN_NAMESPACE {
 			Incrementable<I> &&
 			requires(I& i) {
 				{ --i } -> Same<I>&;
-				i--; Same<decltype(i--), I>;
+				i--; requires Same<I, decltype(i--)>;
 			};
 			// Let a and b be objects of type I.
 			// Axiom: &--a == &a
--- a/include/stl2/detail/iterator/increment.hpp
+++ b/include/stl2/detail/iterator/increment.hpp
@@ -71,10 +71,11 @@ STL2_OPEN_NAMESPACE {
 	template<class I>
 	concept bool WeaklyIncrementable =
 		Semiregular<I> &&
-		requires(I& i) {
+		requires(I i) {
 			typename iter_difference_t<I>;
-			{ ++i } -> Same<I&>&&;
-			i++;
+			requires SignedIntegral<iter_difference_t<I>>;
+			{ ++i } -> Same<I>&; // not required to be equality-preserving
+			i++; // not required to be equality-preserving
 		};
 
 	///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


include/stl2/detail/iterator/increment.hpp
```diff
@@ -96,8 +97,8 @@ STL2_OPEN_NAMESPACE {
 		concept bool Decrementable =
 			Incrementable<I> &&
 			requires(I& i) {
```
CaseyCarter: This is:
```suggestion
			requires(I i) {
```
In P0896. (Not that it makes any difference, but you changed similar occurrences elsewhere.)

include/stl2/detail/iterator/increment.hpp
```diff
@@ -96,8 +97,8 @@ STL2_OPEN_NAMESPACE {
...
-				{ --i } -> Same<I&>&&;
-				{ i-- } -> Same<I>&&;
+				{ --i } -> Same<I>&;
+				i--; Same<decltype(i--), I>;
```
CaseyCarter: Missing `requires` here:
```suggestion
				i--; requires Same<I, decltype(i--)>;
```

include/stl2/detail/range/primitives.hpp
```diff
@@ -42,25 +79,54 @@ STL2_OPEN_NAMESPACE {
...
+		requires SizedSentinel<I, I>
+		constexpr iter_difference_t<I> operator()(I first, I last) const
+		// [[expects axiom: reachable(first, last)]]
+		STL2_NOEXCEPT_RETURN(
+			last - first
+		)
+
+		template<Range R>
+		constexpr iter_difference_t<iterator_t<R>> operator()(R&& r) const
+		noexcept(noexcept(std::declval<__distance_fn const&>()(__stl2::begin(r), __stl2::end(r))))
```
CaseyCarter: Boo GCC for not fixing https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52869. (No change requested.)

CaseyCarter:
> I can't do `advance` yet, as `counted_iterator` provides overloads of `advance`.

You could:
* declare `operator()` overloads for `counted_iterator` in `__advance_fn` and define them in `counted_iterator`'s header, or 
* forward-declare `counted_iterator` and some `__advance(counted_iterator)` overloads here, recognize `counted_iterator`s in `advance` with e.g. `if constexpr (meta::is<I, counted_iterator>)` and dispatch to said overloads. [The first approach is certainly simpler; I suspect this one is ever-so-slightly cheaper at compile time since it doesn't add `operator()` overloads increasing the cost of overload resolution.]

PR #205, opened 2018-12-05
Non slicing subrange

ericniebler:
Attempted fix for ericniebler/stl2#589.

### DIFF
--- a/include/stl2/view/subrange.hpp
+++ b/include/stl2/view/subrange.hpp
@@ -23,6 +23,17 @@
 #include <stl2/view/view_interface.hpp>
 
 STL2_OPEN_NAMESPACE {
+	template<class From, class To>
+	STL2_CONCEPT _ConvertibleToNotSlicing =
+		ConvertibleTo<From, To> &&
+		// A conversion is a slicing conversion if the source and the destination
+		// are both pointers, and if the pointed-to types differ after removing
+		// cv qualifiers.
+		!(_Is<std::decay_t<From>, std::is_pointer> &&
+		  _Is<std::decay_t<To>, std::is_pointer> &&
+		  _NotSameAs<std::remove_pointer_t<std::decay_t<From>>,
+		             std::remove_pointer_t<std::decay_t<To>>>);
+
 	template<class T>
 	STL2_CONCEPT _PairLikeGCCBugs = requires(T t) {
 		{ std::get<0>(t) } -> const std::tuple_element_t<0, T>&;
@@ -36,7 +47,8 @@ STL2_OPEN_NAMESPACE {
 			requires DerivedFrom<std::tuple_size<T>, std::integral_constant<std::size_t, 2>>;
 			typename std::tuple_element_t<0, remove_const_t<T>>;
 			typename std::tuple_element_t<1, remove_const_t<T>>;
-			requires _PairLikeGCCBugs<T>; // Separate named concept to avoid premature substitution.
+			requires _PairLikeGCCBugs<T>; // Separate named concept to avoid
+			                              // premature substitution.
 		};
 
 	template<class T, class U, class V>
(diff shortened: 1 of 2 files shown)


include/stl2/view/subrange.hpp
```diff
@@ -23,6 +23,16 @@
...
 STL2_OPEN_NAMESPACE {
+	template<class From, class To>
+	STL2_CONCEPT _ConvertibleToNotSlicing =
+		ConvertibleTo<From, To> &&
+		// A conversion is a slicing conversion if the source and the destination
+		// are both pointers, and if the pointed-to types differ after removing
+		// cv qualifiers.
+		!(_Is<__uncvref<From>, std::is_pointer> &&
+		  _Is<__uncvref<To>, std::is_pointer> &&
+		  _NotSameAs<std::remove_pointer_t<__uncvref<From>>, std::remove_pointer_t<__uncvref<To>>>);
```
CaseyCarter: The comment says "after removing cv qualifiers":
```suggestion
		  _NotSameAs<remove_cv_t<std::remove_pointer_t<__uncvref<From>>>, remove_cv_t<std::remove_pointer_t<__uncvref<To>>>>);
```

CaseyCarter: Although we usually defer to the core language rules for conversion of pointers-to-array. e.g. for `shared_ptr<T[]>`'s constructors. So maybe:
```suggestion
		  std::is_convertible_v<std::remove_pointer_t<__uncvref<From>>(*)[],
		                        std::remove_pointer_t<__uncvref<To>>(*)[]>);
```

include/stl2/view/subrange.hpp
```diff
@@ -216,6 +241,9 @@ STL2_OPEN_NAMESPACE {
 		}
 	};
 
+	template<Iterator I, Sentinel<I> S>
+	subrange(I, S) -> subrange<I, S>;
+
 	template<Iterator I, Sentinel<I> S>
```
CaseyCarter: Ditto `Sentinel` is redundant. We do need `WeaklyIncrementable` for `iter_difference_t`, but `Iterator` is probably less confusing.
```suggestion
	template<Iterator I, class S>
```

include/stl2/view/subrange.hpp
```diff
@@ -216,6 +241,9 @@ STL2_OPEN_NAMESPACE {
 		}
 	};
 
+	template<Iterator I, Sentinel<I> S>
```
CaseyCarter: These constraints are extraneous - the class template applies them for us.
```suggestion
	template<class I, class S>
```

ericniebler:
> Do you want me to write up the LWG issue or will you do it?

I wouldn't dream of stealing your glory. You can have this one. ;-)

include/stl2/view/subrange.hpp
CaseyCarter: It's a combination of DRY and looking ahead to poor performance of early concepts implementations. For example, clang-concepts compiles the test suite for my libc++ implementation of `<concepts>` two orders of magnitude slower than GCC - 187 seconds vs. 1.7 seconds. This library *is* far more a research tool than a production release, so I won't complain either way.

include/stl2/view/subrange.hpp
```diff
@@ -118,37 +139,44 @@ STL2_OPEN_NAMESPACE {
 
 		subrange() = default;
 
-		constexpr subrange(I i, S s)
-			requires !StoreSize
+		template <_ConvertibleToNotSlicing<I> _I>
+		constexpr subrange(_I i, S s)
```
ericniebler: To be specified as:

```c++
constexpr subrange(_ConvertibleToNotSlicing<I> auto i, S s)
```

Ditto for below.

include/stl2/view/subrange.hpp
```diff
@@ -23,6 +23,17 @@
 #include <stl2/view/view_interface.hpp>
 
 STL2_OPEN_NAMESPACE {
+	template<class From, class To>
+	STL2_CONCEPT _ConvertibleToNotSlicing =
+		ConvertibleTo<From, To> &&
+		// A conversion is a slicing conversion if the source and the destination
+		// are both pointers, and if the pointed-to types differ after removing
+		// cv qualifiers.
+		!(_Is<__uncvref<From>, std::is_pointer> &&
```
timsong-cpp: We need a decay here for `From` and not just `__uncvref`. Consider `pair<Derived(&)[10], some_sentinel>`  used with `_PairLikeConvertibleTo`.

CaseyCarter: I feel I should point out that none of this process is making me any more fond of the conversion from pairs. Tell me again why we want this conversion to work?

ericniebler: I worry also about conversion from things that are _both_ ranges and pair-like, which would be ambiguous today. Yeah, I dunno. I can't think of a really compelling use case right now, except interoperability with old `std::` algorithms like `equal_range` that returns a `std::pair` that represents a range.

CaseyCarter: >  interoperability with old `std::` algorithms like `equal_range` that returns a `std::pair` that represents a range.

I don't know about you, but I've had to explain to people many times why we "don't just make `pair<iterator, iterator>` be a range." It seems like a bad idea to take a step backward and treat them like they denote ranges when we've been explaining to people that an arbitrary pair of iterators is just two iterators with no more semantic meaning. 

(And yes, this is a separate issue.)

CaseyCarter: > I worry also about conversion from things that are both ranges and pair-like, which would be ambiguous today.

We have a `!Range` constraint in the *`pairlike-convertible-XXX`* concepts to disambiguate them.

PR #206, opened 2018-12-07
[alg.search_n] Use "if constexpr" dispatch

CaseyCarter:
... reducing build time for the test from 123 seconds to 3 seconds??!!!?!!

### DIFF
--- a/include/stl2/detail/algorithm/search_n.hpp
+++ b/include/stl2/detail/algorithm/search_n.hpp
@@ -33,13 +33,12 @@
 // search_n [alg.search]
 //
 STL2_OPEN_NAMESPACE {
-	namespace __search_n {
+	struct __search_n_fn {
+	private:
 		template<ForwardIterator I, Sentinel<I> S, class T, class Pred, class Proj>
-		requires
-			IndirectlyComparable<I, const T*, Pred, Proj>
-		I unsized(I first, S last, iter_difference_t<I> count,
-			const T& value, Pred pred, Proj proj)
-		{
+		requires IndirectlyComparable<I, const T*, Pred, Proj>
+		static constexpr I unsized(I first, S last, iter_difference_t<I> count,
+			const T& value, Pred pred, Proj proj) {
 			if (count <= 0) {
 				return first;
 			}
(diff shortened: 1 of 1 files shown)


cjdb:
I literally stopped what I was working on to make sure this wasn't a joke.

tahonermann:
Is it time to start suggesting guidelines like "Don't use concept refinement for overload purposes"?
This should improve error messages for overload resolution failure as well.

CaseyCarter:
> Is it time to start suggesting guidelines like "Don't use concept refinement for overload purposes"?
I'm inclined to say "no" until we can determine *why* this particular example has pathological compile time. All of the cmcstl2 algorithms used pure concept dispatch before there was an `if constexpr` but only 2 of the tests take literally minutes to compile. I need to minimize a perf repro to throw at Jason and Andrew so we can investigate whether this triggers something nasty in the compiler or if it's a fundamental issue with the form of the constraints. Hopefully that will let us give more useful guidance; "don't do this with your constraints" rather than "don't use concept refinement". 

> This should improve error messages for overload resolution failure as well.

Ideally, compilers should prune candidates from the overload set whose constraints subsume the constraints of another non-viable overload before reporting overload resolution failure - much like how today they prune non-viable candidates when there are viable candidates. "Your Foo isn't a Baz" is plenty of information, you don't also need to be told that "Your Foo isn't a FrobnableBaz" as well.

cjdb:
> "Your Foo isn't a Baz" is plenty of information, you don't also need to be told that "Your Foo isn't a FrobnableBaz" as well.

I quite like the idea of

> Your `Foo` isn't a `BidirectionalIterator` because it isn't a `Readable`.

I imagine that would require lots of guidance via attributes.

tahonermann:
> Ideally, compilers should prune candidates from the overload set whose constraints subsume the constraints of another non-viable overload before reporting overload resolution failure

Indeed!  [I wrote something about that a couple of years ago](http://honermann.net/blog/2016/07/18/refining-concepts-improving-error-messages/).  Sadly, ideas come easily, but implementations less so.

CaseyCarter:
"Your `Foo` isn't a `BidirectionalIterator` because it isn't a `Readable`, and it isn't `Readable` because it doesn't have a `*` operator, moron!"

CaseyCarter:
> > reducing build time for the test from 123 seconds to 3 seconds??!!!?!!
> 
> !!!!!!!!!

IKR? #207 reduces `search`'s compile time by a factor of 60.

PR #210, opened 2018-12-07
Implement __niebloid to make niebloids less objectish and more niebloidy

CaseyCarter:


ericniebler:
Killjoy.

CaseyCarter:
> Killjoy.

Most people's experience with Ranges - if any - is from using cmcstl2 on Compiler Explorer. We want them to know what a PITA it is to not be able to pass overload sets as objects, so they'll support EWG papers that enable passing overload sets as objects. And/or support a paper to promote the niebloids to function objects for C++23 if we can't get what we need through EWG.

Also I hate fun.

image

PR #215, opened 2018-12-08
Turns on AddressSanitizer and UndefinedBehaviorSanitizer.

cjdb:
Added .cirrus.yml to enable CI with Docker to facilitate sanitizers.

RTT with Cirrus CI: 191 seconds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,9 @@ cmake_minimum_required(VERSION 3.8)
 
 project(cmcstl2 CXX)
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+find_package(Sanitizer COMPONENTS address undefined REQUIRED)
+
 add_library(stl2 INTERFACE)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ cmake_minimum_required(VERSION 3.8)
 
 project(cmcstl2 CXX)
 
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+find_package(Sanitizer COMPONENTS address undefined REQUIRED)
+
```
CaseyCarter: Don't we need a change to `test/CMakeLists.txt` to link `Sanitizer::all` into the `stl2_test_config` interface library?

cjdb: Yes, I'll add that.

cjdb:
Ninja is cool because when it fails, it automatically blurts out in verbose mode. That doesn't happen in the case of a successful build: would you like me to turn on verbose mode even for successful builds?

PR #221, opened 2018-12-10
more subrange improvements as discussed in ericniebler/stl2#589

ericniebler:


ericniebler:
ATTN: @timsong-cpp

include/stl2/view/subrange.hpp
```diff
@@ -185,20 +173,6 @@ STL2_OPEN_NAMESPACE {
...
-
-		template<_PairLikeConvertibleTo<I, S> PairLike>
-		constexpr subrange(PairLike&& r, iter_difference_t<I> n)
-			requires (K == subrange_kind::sized)
-		: subrange{std::get<0>(static_cast<PairLike&&>(r)),
-			std::get<1>(static_cast<PairLike&&>(r)), n}
-		{}
-
 		template<_NotSameAs<subrange> PairLike>
 		requires _PairLikeConvertibleFrom<PairLike, const I&, const S&>
```
CaseyCarter: Unrelated design discussion: do we really want `subrange<I, S>` to be convertible to e.g. `std::pair<const I&, const S&>` and thereby hand out references to its guts? Or should this instead require `_PairLikeConvertibleFrom<PairLike, I, S>` and return `PairLike(begin(), end())`?

For example, a caller could use such a pair-of-references conversion on a non-const object, validly cast away the constness from the references so obtained, and break the invariant that `distance(first_(), last()) == size_()` when `StoreSize` is `true`. Admittedly, this example seems more Machiavelli than Murphy to me. Thoughts?

include/stl2/view/subrange.hpp
```diff
@@ -140,22 +123,22 @@ STL2_OPEN_NAMESPACE {
 		subrange() = default;
 
 		template <_ConvertibleToNotSlicing<I> _I>
```
CaseyCarter: Eww, I just noticed this parameter name is `_I`. This is not only in the reserved identifier space - which is generally fine - but it's in the `_ALLCAPS` space that implementors typically use for macros. Could we change these to `I_`?

include/stl2/view/subrange.hpp
timsong-cpp: So it remains convertible but produces dangling references instead?

timsong-cpp:
LGTM too.

include/stl2/view/subrange.hpp
CaseyCarter: > So it remains convertible but produces dangling references instead?

`subrange` is a *`forwarding-range`*; the iterator and sentinel in the returned `pair` would have no more potential to dangle than those produced by the non-member `begin` and `end`.

timsong-cpp: "remains convertible to `std::pair<const I&, const S&>`".

ericniebler: > Or should this instead require `_PairLikeConvertibleFrom<PairLike, I, S>` and return `PairLike(begin(), end())`?

Does that really avoid the issue? An `I` is convertible to an `const I&`, no?

CaseyCarter: > Does that really avoid the issue? An `I` is convertible to an `const I&`, no?

The original issue - that people get references to the `subrange`'s guts - is fixed. Tim's followup issue - that the pair-of-references conversion would still succeed and instead reference temporaries - not so much. See my latest post.

PR #226, opened 2018-12-11
fixes for begin/end/cbegin/cend/cdata and subrange/ref_view/iota_view

ericniebler:
IMO, we should spec the rvalue `begin`/`end` overloads for `subrange`, `iota_view`, and `ref_view` to be non-friend function templates, or else as hidden friend function templates using this pattern:

```c++
template<class A, class B> concept same-ish = Same<A const, B const>;

template <...>
struct subrange {
    friend I begin(same-ish<subrange> auto&& x) {
        return x.begin();
    }
};
```

I think this should work, but I can't test it because of the constrained friends gcc bug. With gcc, this overload is ambiguous with the poison-pill overload since the constraints get ignored.

### DIFF
--- a/include/stl2/detail/range/access.hpp
+++ b/include/stl2/detail/range/access.hpp
@@ -109,7 +109,7 @@ STL2_OPEN_NAMESPACE {
 		template<class T> void end(T&&) = delete;
 
 		template<class T>
-		void end(std::initializer_list<T>) = delete;
+		void end(std::initializer_list<T>) = delete; // TODO: file LWG issue
 
 		template<class R>
 		STL2_CONCEPT has_member = std::is_lvalue_reference_v<R> &&
--- a/include/stl2/detail/range/access.hpp
+++ b/include/stl2/detail/range/access.hpp
@@ -26,6 +26,17 @@
 // Range access [range.access]
 //
 STL2_OPEN_NAMESPACE {
+	inline constexpr struct __as_const_fn {
+		template<class T>
+		constexpr const T& operator()(T& t) const noexcept {
+			return t;
+		}
+		template<class T>
+		constexpr const T&& operator()(T&& t) const noexcept {
+			return (T&&)t;
+		}
+	} __as_const {};
+
 	// begin
 	namespace __begin {
 		// Poison pill for std::begin. (See the detailed discussion at
@@ -33,7 +44,7 @@ STL2_OPEN_NAMESPACE {
 		template<class T> void begin(T&&) = delete;
 
 		template<class T>
-		void begin(std::initializer_list<T>&&) = delete;
+		void begin(std::initializer_list<T>) = delete;
 
 		template<class R>
 		STL2_CONCEPT has_member = std::is_lvalue_reference_v<R> &&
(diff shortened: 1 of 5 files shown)


include/stl2/detail/range/access.hpp
```diff
@@ -26,14 +26,25 @@
...
+
 	// begin
 	namespace __begin {
 		// Poison pill for std::begin. (See the detailed discussion at
 		// https://github.com/ericniebler/stl2/issues/139)
 		template<class T> void begin(T&&) = delete;
 
 		template<class T>
-		void begin(std::initializer_list<T>&&) = delete;
+		void begin(std::initializer_list<T>) = delete;
```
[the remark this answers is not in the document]
CaseyCarter:
```suggestion
		void begin(std::initializer_list<T>) = delete; // TODO: file LWG issue
```

include/stl2/detail/range/access.hpp
```diff
@@ -98,7 +109,7 @@ STL2_OPEN_NAMESPACE {
 		template<class T> void end(T&&) = delete;
 
 		template<class T>
-		void end(std::initializer_list<T>&&) = delete;
+		void end(std::initializer_list<T>) = delete;
```
CaseyCarter:
```suggestion
		void end(std::initializer_list<T>) = delete; // TODO: file LWG issue
```

include/stl2/detail/range/access.hpp
```diff
@@ -136,18 +147,10 @@ STL2_OPEN_NAMESPACE {
...
 
-			// Prefer member if it returns Sentinel.
-			template<class R>
-			requires has_member<R&>
-			constexpr auto operator()(R& r) const
-			STL2_NOEXCEPT_RETURN(
-				r.end()
-			)
-
 			// Use ADL if it returns Sentinel.
```
CaseyCarter: Please also strike this comment that I should have struck.

include/stl2/view/ref.hpp
```diff
@@ -24,10 +24,30 @@
...
 	// ref_view [ranges.view.ref]
+	template<Range R>
+	requires std::is_object_v<R>
+	struct ref_view;
+
 	// Not an extension: P1252 makes this user-visible
+	namespace __ref_view_detail {
+		struct __adl_hook {};
+
+		// Not to spec (yet);
```
CaseyCarter:
```suggestion
		// Not to spec: should be hidden friends.
```

CaseyCarter:
> IMO, we should spec the rvalue `begin`/`end` overloads for `subrange`, `iota_view`, and `ref_view` to be non-friend function templates, or else as hidden friend function templates using this pattern:

I agree, except for `ref_view` which *should* be passed by value in a register.

ericniebler:
> I agree, except for `ref_view` which _should_ be passed by value in a register.

In that case, it should be:

```c++
template<class R>
struct ref_view {
  friend iterator_t<R> begin(Same<ref_view> auto r) { return ranges::begin(r); }
...
};
```

and not a non-template friend, which gets us back into hot water with conversions and constraint recursion.

Maybe I should try removing the `view_interface` conversion-to-container and see if it improves things.

PR #228, opened 2018-12-12
Remove superfluous and problematic view converting constructors; fixes #223

### LINKED ISSUE #223: View Pipe Form: Mixing reverse and filter
Continuing from #214.

This works:

```cpp
#include <experimental/ranges/algorithm>
#include <experimental/ranges/iterator>
#include <experimental/ranges/ranges>
#include <algorithm>
#include <iostream>
#include <vector>

int main()
{
    namespace ranges = std::experimental::ranges;
    int a[] = {1, 7, 3, 6, 5, 2, 4, 8};
    auto r = a
        | ranges::view::filter([](int i) { return i % 2 == 0; })
        | ranges::view::reverse;
    ranges::copy(a, ranges::ostream_iterator<int>(std::cout, " "));
    std::cout << std::endl;
    ranges::copy(r, ranges::ostream_iterator<int>(std::cout, " "));
    std::cout << std::endl;
}
```

Reversing the two view adapters doesn't. Neither order works if I change the type of `a` from `int [8]` to `std::vector<int>`. All three failure cases have different error messages.

ericniebler:


CaseyCarter:
What are your thoughts on the "perform implicit conversions in the callee" pattern:
```c++
template<View V>
struct adaptor {
  V v_;

  adaptor() = default;
  template<ConvertibleTo<V> T>
  adaptor(T&& t) : V(static_cast<V>(std::forward<T>(t))) {}
};

template<class R>
adaptor(R&&) -> adaptor<all_view<R>>;
```
which, as far as I can see, is equivalent except for move constructing and destroying one less `V` when initialized with an expression of type `V`?

CaseyCarter:
This change removes all uses of `_ConstructibleFromRange` but doesn't remove `_ConstructibleFromRange`.

include/stl2/view/transform.hpp
```diff
@@ -41,11 +41,6 @@ STL2_OPEN_NAMESPACE {
 		constexpr transform_view(R base, F fun)
 		: base_(std::move(base)), fun_(std::move(fun)) {}
 
-		template<InputRange O>
-		requires ViewableRange<O> && _ConstructibleFromRange<R, O>
```
CaseyCarter: This change removes all uses of `_ConstructibleFromRange` but doesn't remove `_ConstructibleFromRange`.

ericniebler:
I have a mild preference for unconstrained non-templates if that has the right semantics. I'm not overly concerned with an extra move of a view.

PR #233, opened 2018-12-13
Added associated types for ranges (P1035)

cjdb:
Adds:

* `range_value_t`
* `range_difference_t`
* `range_reference_t`
* `range_rvalue_reference_t`

### DIFF
--- a/include/stl2/detail/range/concepts.hpp
+++ b/include/stl2/detail/range/concepts.hpp
@@ -159,6 +159,20 @@ STL2_OPEN_NAMESPACE {
 	STL2_CONCEPT ViewableRange =
 		Range<Rng> &&
 		(_RangeImpl<Rng> || View<Rng>);
+
+	namespace ext {
+		template<Range R>
+		using range_value_t = iter_value_t<iterator_t<R>>;
+
+		template<Range R>
+		using range_difference_t = iter_difference_t<iterator_t<R>>;
+
+		template<Range R>
+		using range_reference_t = iter_reference_t<iterator_t<R>>;
+
+		template<Range R>
+		using range_rvalue_reference_t = iter_reference_t<iterator_t<R>>;
+	} // namespace ext
 } STL2_CLOSE_NAMESPACE
 
 #endif


include/stl2/detail/range/concepts.hpp
```diff
@@ -159,6 +159,20 @@ STL2_OPEN_NAMESPACE {
 	STL2_CONCEPT ViewableRange =
 		Range<Rng> &&
 		(_RangeImpl<Rng> || View<Rng>);
+
+	namespace ext {
```
CaseyCarter: These were approved by LEWG, right? If so, let's put them directly in `__stl2` and not bother with `ext`.

include/stl2/detail/range/concepts.hpp
```diff
@@ -159,6 +159,20 @@ STL2_OPEN_NAMESPACE {
...
+		template<Range R>
```
CaseyCarter: Design question: Why constrain these with `Range` instead of constraining `iterator_t` and `sentinel_t`?

include/stl2/detail/range/concepts.hpp
cjdb: I think everything in P1035 is going back to LEWG for confirmation.

CaseyCarter: Yes, I misremembered the minutes. `ext` it is.

include/stl2/detail/range/concepts.hpp
CaseyCarter: Per personal discussion on Slack, @cjdb is going to add this suggestion to P1035.

cjdb: https://www.cjdb.com.au/wg21/ranges/d1035/d1035.html#introducing-associated-types-for-ranges

cjdb: Reflected here: https://www.cjdb.com.au/wg21/ranges/d1035/d1035.html#amendment-to-the-associated-iterator-types

ericniebler: I didn't realize that the Better Safer CPO paper made this possible. Very nice! (y)

PR #234, opened 2018-12-13
Generate view

cjdb:
Added `generate_view`.


include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+			__iterator(generate_view& v)
+				: view_{&v}
+			{}
+
+			constexpr decltype(auto) operator*() noexcept
+			requires (!Invocable<F const>)
+			{ return deref_impl(*this); }
+
+			constexpr decltype(auto) operator*() const noexcept
```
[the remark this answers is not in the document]
CaseyCarter: @ericniebler range-v3's `generate_view` is similar. We need to rethink.

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+			using iterator_concept = input_iterator_tag;
+
+			__iterator() = default;
+
+			__iterator(generate_view& v)
+				: view_{&v}
+			{}
+
+			constexpr decltype(auto) operator*() noexcept
+			requires (!Invocable<F const>)
```
CaseyCarter:
```suggestion
			requires (!Invocable<const F&>)
```

CaseyCarter: It's somewhat silly that this library is `const` west to agree with the working draft's style when all of the maintainers prefer east `const`. I may revisit that decision, but for now let's keep things consistent.

CaseyCarter: Also, we should probably require `Same<invoke_result_t<F&>, invoke_result_t<const F&>>` as part of our choice to call `const`-qualified `F`.

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+			{ return deref_impl(*this); }
+
+			constexpr decltype(auto) operator*() const noexcept
+			requires Invocable<F const>
```
CaseyCarter: Ditto `const F&`, ditto missing `Same` requirement.

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+			constexpr __iterator& operator++() noexcept
+			{
+				view_->as_cache().reset();
+				return *this;
+			}
+
+			constexpr __iterator operator++(int) noexcept
+			{
+				auto result = *this;
```
CaseyCarter: It's somewhat silly to store a copy when pre-increment doesn't modify the iterator ;)

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+			constexpr decltype(auto) operator*() noexcept
+			requires (!Invocable<F const>)
+			{ return deref_impl(*this); }
+
+			constexpr decltype(auto) operator*() const noexcept
+			requires Invocable<F const>
+			{ return deref_impl(*this); }
+
+			constexpr __iterator& operator++() noexcept
```
CaseyCarter: `std::optional::reset` is not `constexpr`.
```suggestion
			__iterator& operator++() noexcept
```

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+			{
+				view_->as_cache().reset();
+				return *this;
+			}
+
+			constexpr __iterator operator++(int) noexcept
```
CaseyCarter:
```suggestion
			__iterator operator++(int) noexcept
```

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+			{ return !(y == x); }
+		private:
+			generate_view* view_;
+
+			template<class Self>
+			static constexpr decltype(auto) deref_impl(Self&& self)
+			{
+				auto& view = *self.view_;
+				if (!view.as_cache()) {
+					view.as_cache() = view.as_box().get()();
```
CaseyCarter: Since we know the cache is empty, we should use `emplace` here:
```suggestion
					view.as_cache().emplace(view.as_box().get()());
```

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+			constexpr friend bool operator!=(__iterator x, unreachable y) noexcept
+			{ return !(x == y); }
+
+			constexpr friend bool operator!=(unreachable x, __iterator y) noexcept
+			{ return !(y == x); }
+		private:
+			generate_view* view_;
+
+			template<class Self>
+			static constexpr decltype(auto) deref_impl(Self&& self)
```
CaseyCarter:
```suggestion
			static auto&& deref_impl(Self&& self)
```

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+			using iterator_category = input_iterator_tag;
+			using iterator_concept = input_iterator_tag;
+
+			__iterator() = default;
+
+			__iterator(generate_view& v)
+				: view_{&v}
+			{}
+
+			constexpr decltype(auto) operator*() noexcept
```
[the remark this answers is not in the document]
CaseyCarter:
```suggestion
			constexpr auto&& operator*() noexcept
```

include/stl2/view/generate_view.hpp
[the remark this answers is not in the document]
CaseyCarter:
```suggestion
			constexpr auto&& operator*() const noexcept
```

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+// a simple type to test move semantics
+struct MoveOnlyString {
+	char const* sz_;
+
+	MoveOnlyString(char const* sz = "")
+	  : sz_(sz)
+	{}
+	MoveOnlyString(MoveOnlyString&& that)
+	  : sz_(that.sz_)
```
CaseyCarter:
```suggestion
	  : sz_(ranges::exchange(that.sz_, ""))
```

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+	{
+		that.sz_ = "";
```
CaseyCarter:
```suggestion
```

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+	}
+	MoveOnlyString(MoveOnlyString const&) = delete;
+	MoveOnlyString& operator=(MoveOnlyString&& that)
+	{
+		sz_ = that.sz_;
```
CaseyCarter:
```suggestion
		sz_ = ranges::exchange(that.sz_, "");
```

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+		that.sz_ = "";
```
CaseyCarter:
```suggestion
```

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+	friend std::ostream& operator<<(std::ostream& sout, MoveOnlyString const& str)
+	{ return sout << '"' << str.sz_ << '"'; }
+};
+
+struct MoveOnlyFunction {
+	MoveOnlyString str_;
+	int i_;
+
+	char operator()()
+	{ return str_.sz_[i_++]; }
```
CaseyCarter: I'm concerned that this doesn't explode *very loudly* if asked to access past-the-end of the string.

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+};
+
+int main()
+{
+	// Test for constant generator functions
+	{
+		int i = 0, j = 1;
+		auto fib = view::generate([&]()->int{int tmp = i; i += j; std::swap(i, j); return tmp;});
```
CaseyCarter:
```suggestion
		auto fib = view::generate([&]{ int tmp = i; i += j; std::swap(i, j); return tmp; });
```

CaseyCarter: (Meh, I don't actually care for test code, just pointing out the `() -> int` is extraneous.)

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+		auto fib = view::generate([=]()mutable->int{int tmp = i; i += j; std::swap(i, j); return tmp;});
+		static_assert(ranges::InputRange<decltype(fib)>);
+		static_assert(ranges::View<decltype(fib)>);
+		CHECK_EQUAL(fib | view::take_exactly(10), {0,1,1,2,3,5,8,13,21,34});
+		// The generator cannot be called when it's const-qualified, so "fib const"
+		// does not model View.
+		static_assert(!ranges::View<decltype(fib) const>);
+	}
+
+	// Test for move-only generator functions
```
CaseyCarter: I'd prefer wrapping this block with `#if 0 // cmcstl2 doesn't support move-only input views` to commenting it out.

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+		CHECK(bool(*i == MoveOnlyString{"hi"}));
+		CHECK(bool(*r.begin() == MoveOnlyString{"hi"}));
+		CHECK(bool(*r.begin() == MoveOnlyString{"hi"}));
+		static_assert(ranges::InputRange<decltype(r)>);
+		static_assert(ranges::View<decltype(r)>);
+		CHECK_EQUAL(r, {MoveOnlyString{"hi"}, MoveOnlyString{"ii"}});
+		static_assert(std::is_same<ranges::ext::range_reference_t<decltype(r)>, MoveOnlyString&>::value, "");
+	}
+
+	// // Test for generator functions that return internal references
```
CaseyCarter:
```suggestion
	// Test for generator functions that return internal references
```

test/view/generate_view.cpp
```diff
@@ -0,0 +1,132 @@
...
+	// // https://github.com/ericniebler/range-v3/issues/807
```
CaseyCarter:
```suggestion
	// https://github.com/ericniebler/range-v3/issues/807
```

CaseyCarter:
@cjdb You probably want to read the "This design is broken and needs a rethink" comment before investing any work in the others.

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+			}
+
+			constexpr __iterator operator++(int) noexcept
+			{
+				auto result = *this;
+				++(*this);
+				return result;
+			}
+
+			constexpr friend bool operator==(__iterator, unreachable) noexcept
```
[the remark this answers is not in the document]
cjdb: TIL this isn't necessary.

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+									, private detail::non_propagating_cache<invoke_result_t<F&>> {
+		private:
+			using result_t = invoke_result_t<F&>;
+
+			template<bool IsConst>
+			struct __iterator;
+		public:
+			generate_view() = default;
+
+			constexpr generate_view(F f) noexcept(is_nothrow_move_constructible_v<F>)
```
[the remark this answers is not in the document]
cjdb: Yes. Re `std` qualifications: I assume this is a general rule to apply from now on?

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
...
+		struct generate_view<F>::__iterator {
+			using value_type = result_t;
+			using difference_type = ptrdiff_t; // I guess?
+			using iterator_category = input_iterator_tag;
+			using iterator_concept = input_iterator_tag;
+
+			__iterator() = default;
+
+			__iterator(generate_view& v)
+				: view_{&v}
```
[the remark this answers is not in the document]
cjdb: > I'll hold my nose and merge if you don't want to change it
Nah, it helps reinforce 'good library implementation' if you flag this stuff 😄

include/stl2/view/generate_view.hpp
cjdb: This is an interesting problem that SFINAE causes (at least, I don't _think_ this one is a GCC bug):

```
auto&& operator*() const
requires (Invocable<const F&>)
{
   static_assert(Same<result_t, invoke_result_t<const F&>>);
   return deref_impl(*this);
}
```

include/stl2/view/generate_view.hpp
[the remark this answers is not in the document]
CaseyCarter: Yes. There are still quite a few violations - I haven't really audited - but ideally the library will continue to work if someone edits the definitions of `STL2_(OPEN|CLOSE)_NAMESPACE` and `__stl2` to some other namespace that isn't nested within `std`.

include/stl2/view/generate_view.hpp
CaseyCarter: I think I'm missing some context - what's the problem?

cjdb: I was mostly complaining about the fact that `invoke_result_t<const F&>` is evaluated before `Invocable<const F&>`, and thus an instantiation error occurs.

After five minutes of thinking, I realised it could be done using either _requires-clause_(s) or a concept (which is what I chose).

test/view/generate_view.cpp
cjdb: FYI this will simplify a lot of my move constructors from now on!

cjdb:
Here's a slightly different `generate` for side-by-side comparison: https://github.com/cjdb/cmcstl2/pull/22

I'm pretty confident that it could have been commit to this branch, but I wanted to diff the changes (I like the GitHub interface for diffs).

include/stl2/view/generate_view.hpp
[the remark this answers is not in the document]
CaseyCarter: > I _think_ the path forward here is to _not_ require `*` to be valid for `const` `Readable`s, but to instead require it to be non-modifying (in the sense of [concepts.equality]).

This has lots of scary ramifications for the rest of the design, for example, all of the iterator adaptors have `iter_move` and `iter_swap` overloads that accept `const` references - those would need to change to forwarding references.

include/stl2/view/generate_view.hpp
CaseyCarter: If `requires Invocable<F const&> && Same<result_t, invoke_result_t<F const&>>` doesn't work because GCC substitutes into the `invoke_result_t` when `Invocable` isn't satisfied, that's a bug.

test/view/generate_view.cpp
CaseyCarter: Yeah, I use `std::exchange(foo, nullptr)` *all* the time.

test/view/generate_view.cpp
cjdb: I hope `STL2_EXPECT` is loud enough (I can also make this change upstream if you like... there's a `RANGES_EXPECT`, right?).

CaseyCarter: > I hope `STL2_EXPECT` is loud enough 

Yes, it is.

>(I can also make this change upstream if you like... there's a `RANGES_EXPECT`, right?).

That would be nice. And yes there is, and I strongly suspect that they have the same behavior. ;)

include/stl2/view/generate_view.hpp
```diff
@@ -0,0 +1,130 @@
+#ifndef STL2_VIEW_GENERATE_VIEW_HPP
```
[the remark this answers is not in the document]
CaseyCarter: You removed `VIEW` from the header guard, which is great, but it's still in the name of the header. Also this header needs to be added to `stl2/range.hpp`.

test/move_only_string.hpp
```diff
@@ -0,0 +1,78 @@
...
+		, length_(std::strlen(sz))
+		{}
+		move_only_string(move_only_string&& that) noexcept
+		: sz_(ranges::exchange(that.sz_, ""))
+		, length_(ranges::exchange(that.length_, 0))
+		{}
+		move_only_string(move_only_string const&) = delete;
+		move_only_string& operator=(move_only_string&& that) noexcept
+		{
+			*this = ranges::exchange(that, move_only_string{});
```
CaseyCarter: No change requested, just pointing out for future reference that `exchange`'s second template parameter defaults to the first, so you could equivalently write `*this = ranges::exchange(that, {});` here. `exchange(foo, {})` is very nicely idiomatic for "put `foo` in a value-initialized state and return its prior state".

test/view/generate_view.cpp
```diff
@@ -0,0 +1,113 @@
...
+		auto fib = view::generate([=]()mutable->int{int tmp = i; i += j; std::swap(i, j); return tmp;});
+		static_assert(ranges::InputRange<decltype(fib)>);
+		static_assert(ranges::View<decltype(fib)>);
+		CHECK_EQUAL(fib | view::take_exactly(10), {0,1,1,2,3,5,8,13,21,34});
+		// The generator cannot be called when it's const-qualified, so "fib const"
+		// does not model View.
+		static_assert(!ranges::View<decltype(fib) const>);
+	}
+
+#if 0 // cmcstl2 doesn't support move-only generator functions
```
CaseyCarter:
```suggestion
#if 0 // unlike range-v3, cmcstl2 doesn't support move-only input views
```

CaseyCarter:
> There's 25 commits atm, might be worthwhile rebasing and squashing before merging. I can do this now or when you're ready to merge?

If you're happy with a simple squash, I'll do so when I merge. If you want any kind of more detailed commit history, that's on you ;)

include/stl2/view/generate.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include <stl2/detail/semiregular_box.hpp>
+#include <stl2/detail/view/view_closure.hpp>
+#include <stl2/view/view_interface.hpp>
+#include <type_traits>
+#include <utility>
+
+STL2_OPEN_NAMESPACE {
+	namespace ext {
+		template<class F>
+		STL2_CONCEPT __ConstInvocable =
```
CaseyCarter:
```suggestion
		META_CONCEPT __ConstInvocable =
```

cjdb:
> If you're happy with a simple squash, I'll do so when I merge. If you want any kind of more detailed commit history, that's on you ;)

Ugh, I missed this one.

include/stl2/view/generate_view.hpp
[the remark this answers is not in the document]
cjdb: Unresolving this to confirm that it has actually been addressed (I think one of my pushes resolved it).

Issue #169, opened 2018-07-15
meta from cmcstl2 clashes with meta from range-v3


cjdb:
```cpp
// config 1 fails
#include <something from cmcstl2>
#include <range/v3/all.hpp>

// config 2 works
#include <range/v3/all.hpp>
#include <something from cmcstl2>
```

Diagnostic sample from config 1:

```
In file included from ../range-v3/include/range/v3/view.hpp:21,
                 from ../range-v3/include/range/v3/all.hpp:21,
                 from ../ltcpp/test/ltcpp/ast/expression/../../common.hpp:23,
                 from ../ltcpp/test/ltcpp/ast/expression/expression_common.hpp:22,
                 from ../ltcpp/test/ltcpp/ast/expression/array_expression_pass.cpp:17:
../range-v3/include/range/v3/view/adjacent_filter.hpp:67:36: error: ‘const_if_c’ in namespace ‘meta’ does not name a template type
                 using Base = meta::const_if_c<const_iterable, Rng>;
                                    ^~~~~~~~~~
../range-v3/include/range/v3/view/adjacent_filter.hpp:68:38: error: ‘const_if_c’ in namespace ‘meta’ does not name a template type
                 using Parent = meta::const_if_c<const_iterable, adjacent_filter_view>;
                                      ^~~~~~~~~~
../range-v3/include/range/v3/view/adjacent_filter.hpp:69:17: error: ‘Parent’ does not name a type; did you mean ‘Pred’?
                 Parent *rng_;
                 ^~~~~~
                 Pred
../range-v3/include/range/v3/view/adjacent_filter.hpp:72:41: error: expected ‘)’ before ‘&’ token
                 constexpr adaptor(Parent &rng) noexcept
                                  ~      ^~
                                         )
../range-v3/include/range/v3/view/adjacent_filter.hpp:75:61: error: ‘Base’ was not declared in this scope
                 RANGES_CXX14_CONSTEXPR void next(iterator_t<Base> &it) const
                                                             ^~~~
```

This goes on for quite some time. Is it possible to sync cmcstl2's meta library with range-v3's, or have there been customisations along the way?

CaseyCarter:
[...]

PR #238, opened 2018-12-15
Reduce meta/stl2 duplication:

CaseyCarter:
meta:
* Define macro `META_CONCEPT_BARRIER` to wrap atomic expressions in a `constexpr bool` to avoid inadvertent subsumption when using TS concepts.

* Replace implementation of `meta::Same` with that of `__stl2::_SameImpl`.

* Add `meta::is_v` variable template form of `meta::is`. Use it to implement `meta::List` and `meta::is`.

* Update definition of `meta::Integral`, now that https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68434 is fixed. Guard against substitutions producing non-constant expressions.

stl2:
* Replace `_Valid` with `meta::_Valid`

* Replace `STL2_CONCEPT` with `META_CONCEPT`

* Removed unused `_Is` and `_IsNot` concepts

* Use `meta::Same` to implement `Same`, `_Decayed`, `_OneOf`, and `_NotSameAs`.

* Remove unused `__as_integer_sequence` alias.

* Factor out intrinsic selection for `is_convertible` into a variable template, use it in `ConvertibleTo`, `DerivedFrom`, `is_nothrow_convertible_v`, and `__decay_copy`.

### DIFF
--- a/include/stl2/detail/algorithm/sample.hpp
+++ b/include/stl2/detail/algorithm/sample.hpp
@@ -24,7 +24,7 @@
 STL2_OPEN_NAMESPACE {
 	namespace __sample {
 		template<class I, class S, class O, class Gen>
-		STL2_CONCEPT constraint =
+		META_CONCEPT constraint =
 			InputIterator<I> && Sentinel<S, I> && WeaklyIncrementable<O> &&
 			IndirectlyCopyable<I, O> &&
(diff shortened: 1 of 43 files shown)


CaseyCarter:
@ericniebler make sure you're happy with the changes to `meta`; I plan to merge the three different `meta` streams together to fix #169.

ericniebler:
Looks good to me.

PR #246, opened 2018-12-22
A set of updates

cjdb:


include/stl2/detail/algorithm/includes.hpp
```diff
@@ -21,48 +21,43 @@
...
-		return __stl2::includes(
-			begin(rng1), end(rng1),
-			begin(rng2), end(rng2),
-			__stl2::ref(comp),
-			__stl2::ref(proj1),
-			__stl2::ref(proj2));
-	}
+		template<InputRange R1, InputRange R2, class Proj1 = identity, class Proj2 = identity,
+			IndirectStrictWeakOrder<projected<iterator_t<R1>, Proj1>,
+				                     projected<iterator_t<R2>, Proj2>> Comp = less>
```
cjdb: This is tabbed (for indentation) and spaced so that it's aligned with the `projected` above, but it looks weird here. Should these spaces be removed?

CaseyCarter: It's weird because there are three tabs on 50 and four tabs on 51. It will align properly if there are the same number of tabs on both lines and a space on the second line for every character in `IndirectStrictWeakOrder<`.

include/stl2/detail/algorithm/set_intersection.hpp
```diff
@@ -22,50 +22,48 @@
...
+					++result;
+					++first1;
+					++first2;
+				}
 			}
+			return {std::move(first1), std::move(first2), std::move(result)};
+		}
+
+		template<InputRange R1, InputRange R2, WeaklyIncrementable O,
+				class Comp = less, class Proj1 = identity, class Proj2 = identity>
```
CaseyCarter:
```suggestion
			class Comp = less, class Proj1 = identity, class Proj2 = identity>
```

include/stl2/detail/algorithm/includes.hpp
CaseyCarter: Yes, this is annoyingly fiddly, which is why I mostly just add an indent on continuation lines and don't bother with alignment much. If you *really* want them aligned but don't want to bother with spaces, you could add an extra wrap:
```c++
\t\t\tIndirectStrictWeakOrder<
\t\t\t\tprojected</*...*/>,
\t\t\t\tprojected</*...*/>>
```

PR #252, opened 2018-12-23
Copymove

CaseyCarter:


ericniebler:
For the record, I will not be niebloidifying range-v3. Support for higher-ordered programming is a feature.

PR #259, opened 2018-12-28
Wip

CaseyCarter:


CaseyCarter:
This PR completes the transition of the algorithms from function templates with concept overloading to niebloids with `if constexpr` dispatch. Total CPU time for a test suite build on my home machine has gone down from 22:40 to 7:30.

PR #261, opened 2018-12-29
changes istream_view requirements and adds test

cjdb:
`istream_view` previously required its type to model `Semiregular`. This commit weakens the requirement to `Movable` and adds a test (from range-v3) to prove its correctness.


include/stl2/view/istream.hpp
```diff
@@ -19,21 +19,24 @@
 #include <stl2/detail/raw_ptr.hpp>
 #include <stl2/detail/semiregular_box.hpp>
 #include <stl2/detail/iostream/concepts.hpp>
+#include <stl2/detail/iterator/default_sentinel.hpp>
 #include <stl2/view/view_interface.hpp>
 
 STL2_OPEN_NAMESPACE {
 	namespace ext {
-		template<Semiregular Val>
+		template<Movable Val>
```
CaseyCarter: This is somewhat problematic in that `istream_view<T>` is a `Range` but not a `View` if `T` doesn't model `Copyable`. Note that range-v3 implements a relaxation of `View` which allows move-only single-pass views and requires `Copyable` only for forward views. We need to propose either that semantic or an even weaker relaxation that doesn't require copyability for *any* views, and we need to get that done in time for C++20. (I could care less if `istream_view` supports move-only types, but enabling coroutine generators to participate in view composition is crucial.)

CaseyCarter: Clarifying: I'll merge this as-is, but we need the relaxation proposal in hand rather than proposing an `XXX_view` type for C++20 that isn't always a view.

CaseyCarter: Oops: We need to require `DefaultConstructible<Val>` so we can construct an object into which to extract a value. (Addressed in https://github.com/cjdb/cmcstl2/pull/24.)

include/stl2/view/istream.hpp
```diff
@@ -49,7 +52,7 @@ STL2_OPEN_NAMESPACE {
 			{ return {}; }
 		};
 
-		template<Semiregular Val>
+		template<class Val>
```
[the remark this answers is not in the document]
CaseyCarter: (Addressed in https://github.com/cjdb/cmcstl2/pull/24.)

include/stl2/view/istream.hpp
cjdb: > We need to propose either that semantic or an even weaker relaxation that doesn't require copyability for any views, and we need to get that done in time for C++20.

I'm more comfortable looking into this relaxation, rather than going with a blanket weakening. Does this need to be done by Kona?

CaseyCarter: > Does this need to be done by Kona?

It's a change to an existing feature in the WD, rather than a new feature proposal, so I think we could do it whenever.

PR #271, opened 2019-01-09
reference_wrapper should be Copyable, eh?

CaseyCarter:
Also implement LWG2993 for the reference_wrapper-alike in the iterator test


include/stl2/detail/functional/invoke.hpp
```diff
@@ -82,7 +82,7 @@ STL2_OPEN_NAMESPACE {
 	template<__can_reference T>
 	struct reference_wrapper {
 	private:
-		T& t_;
+		T* t_;
```
CaseyCarter: I must have been drunk when I implemented this??!

PR #274, opened 2019-01-26
Add single_ref_view to ext

ghost:
@CaseyCarter @cjdb here's an implementation of the `single_ref_view` idea I bounced around in the slack channel earlier today. The implementation strategy basically follows that of `rev_view`, but the adapter behaves like `single_view`.

It achieves basically what you'd get from `single_view{std::ref(obj)}`, but in a first-class way without the `reference_wrapper`! I don't know how widely applicable this is, but at the moment I'm pretty sure I have a good use-case.

Snippet demonstrating the main difference between this and `single_view`:
```
{
    auto obj = T(...);
    auto v = ranges::view::single(obj);
    assert(&obj != &*v.begin());
}

{
    auto obj = T(...);
    auto v = ranges::view::single_ref(obj);
    assert(&obj == &*v.begin());
}
```


include/stl2/view/single_ref.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <stl2/detail/fwd.hpp>
+#include <stl2/detail/concepts/object.hpp>
+#include <stl2/detail/view/view_closure.hpp>
+#include <stl2/view/view_interface.hpp>
+
+STL2_OPEN_NAMESPACE {
+namespace ext {
+	template<class T>
+	requires std::is_object_v<T>
+	class single_ref_view : public view_interface<single_ref_view<T>> {
```
cjdb: s/`class`/`struct`/

For future reference, in cmcstl2:
1. `class` means template parameter
2. `struct` means user-defined type
3. `typename` means dependent type

include/stl2/view/single_ref.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+	template <class T>
+	single_ref_view(T&) -> single_ref_view<T>;
+}
+
+	namespace view::ext {
+		struct __single_ref_fn : detail::__pipeable<__single_ref_fn> {
+			template<class T>
+			constexpr auto operator()(T&& t) const
+			STL2_REQUIRES_RETURN(
```
cjdb: This doesn't throw: it can be `STL2_NOEXCEPT_REQUIRES_RETURN`.

include/stl2/view/single_ref.hpp
```diff
@@ -0,0 +1,74 @@
...
+#include <utility>
+
+#include <stl2/type_traits.hpp>
+#include <stl2/detail/fwd.hpp>
+#include <stl2/detail/concepts/object.hpp>
+#include <stl2/detail/view/view_closure.hpp>
+#include <stl2/view/view_interface.hpp>
+
+STL2_OPEN_NAMESPACE {
+namespace ext {
```
cjdb: Can this namespace be indented too please?

include/stl2/view/single_ref.hpp
```diff
@@ -0,0 +1,74 @@
...
+		requires requires(U&& u) { fun(static_cast<U&&>(u)); }
+#endif
+		constexpr explicit single_ref_view(U&& u)
+		noexcept(noexcept(fun(static_cast<U&&>(u)))) // strengthened
+		: ref_{std::addressof(static_cast<T&>(static_cast<U&&>(u)))} {}
+
+		constexpr T* begin() const noexcept { return data(); }
+		constexpr T* end() const noexcept { return data() + 1; }
+		constexpr static std::ptrdiff_t size() noexcept { return 1; }
+		constexpr T* data() const noexcept { return ref_; }
```
cjdb: There should be both a `const` and a non-`const` overload, returning a pointer-to-`const` and a pointer, respectively.

include/stl2/view/single_ref.hpp
```diff
@@ -0,0 +1,74 @@
...
+#else
+		template<_NotSameAs<single_ref_view> U>
+		requires requires(U&& u) { fun(static_cast<U&&>(u)); }
+#endif
+		constexpr explicit single_ref_view(U&& u)
+		noexcept(noexcept(fun(static_cast<U&&>(u)))) // strengthened
+		: ref_{std::addressof(static_cast<T&>(static_cast<U&&>(u)))} {}
+
+		constexpr T* begin() const noexcept { return data(); }
+		constexpr T* end() const noexcept { return data() + 1; }
```
cjdb: There should be both a `const` and a non-`const` overload, returning a pointer-to-`const` and a pointer, respectively.

include/stl2/view/single_ref.hpp
```diff
@@ -0,0 +1,74 @@
...
+			requires(U&& u) { fun(static_cast<U&&>(u)); }
+#else
+		template<_NotSameAs<single_ref_view> U>
+		requires requires(U&& u) { fun(static_cast<U&&>(u)); }
+#endif
+		constexpr explicit single_ref_view(U&& u)
+		noexcept(noexcept(fun(static_cast<U&&>(u)))) // strengthened
+		: ref_{std::addressof(static_cast<T&>(static_cast<U&&>(u)))} {}
+
+		constexpr T* begin() const noexcept { return data(); }
```
cjdb: There should be both a `const` and a non-`const` overload, returning a pointer-to-`const` and a pointer, respectively.

cjdb (changes requested):
Overall: LGTM.

include/stl2/view/single_ref.hpp
cjdb: I suspect that @CaseyCarter will quote [[res.on.data.races]](http://eel.is/c++draft/res.on.data.races), and it makes me queasy when I see a `const` member function that returns a mutable pointer.

cjdb (approved):

PR #279, opened 2019-03-09
Cmake adjustments for cmcstl2 as a subproject

daltonwoodard:
Here's some minor adjustments to the CMake files to allow cmcstl2 to play nicely as a subproject.


CMakeLists.txt
```diff
@@ -14,6 +14,14 @@ cmake_minimum_required(VERSION 3.8)
 
 project(cmcstl2 CXX)
 
+# check if cmcstl2 is a subproject
+if(CMAKE_CURRENT_LIST_DIR STREQUAL CMAKE_SOURCE_DIR)
+    set(STL2_NOT_SUBPROJECT ON)
+endif()
+
+option(STL2_BUILD_EXAMPLES "Build stl2 examples" ON)
+option(STL2_BUILD_TESTING "Build stl2 tests" ON)
```
CaseyCarter: Should these default to `STL2_NOT_SUBPROJECT` instead of `ON`?

CMakeLists.txt
```diff
@@ -36,8 +44,12 @@ install(
     FILES ${PROJECT_BINARY_DIR}/cmcstl2-config.cmake
     DESTINATION lib/cmake/cmcstl2)
 
-add_subdirectory(examples)
+if(STL2_NOT_SUBPROJECT OR STL2_BUILD_EXAMPLES)
```
CaseyCarter: Why test `STL2_NOT_SUBPROJECT` here and on 51 rather than testing only `STL2_BUILD_foo`?

CMakeLists.txt
```diff
@@ -14,6 +14,14 @@ cmake_minimum_required(VERSION 3.8)
 
 project(cmcstl2 CXX)
 
+# check if cmcstl2 is a subproject
+if(CMAKE_CURRENT_LIST_DIR STREQUAL CMAKE_SOURCE_DIR)
+    set(STL2_NOT_SUBPROJECT ON)
```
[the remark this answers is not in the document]
daltonwoodard: Certainly! Will do.

daltonwoodard:
I got rid of the `STL2_NOT_SUBPROJECT` variable entirely since it wasn't needed elsewhere, and I made the default option values for the examples and tests builds conditional on the former subproject check.

CaseyCarter:
Thanks!

PR #280, opened 2019-03-23
[WIP] Implements P1207

cor3ntin:
This implements move-only iterators.
I still have to write tests for 35 or so more
algorithms accepting InputIterator.

### DIFF
--- a/include/stl2/detail/iterator/basic_iterator.hpp
+++ b/include/stl2/detail/iterator/basic_iterator.hpp
@@ -81,14 +81,14 @@ STL2_OPEN_NAMESPACE {
 		// is an opt-out trait that tells basic_iterator that an Input cursor is NOT
 		// Forward despite having equality.
 		template<class>
-		constexpr bool single_pass = false;
+		constexpr bool single_pass = true;
 		template<class C>
-		requires
-			requires {
+		requires (copyable<C> &&
+			!requires {
 				typename C::single_pass;
 				requires bool(C::single_pass::value);
-			}
-		constexpr bool single_pass<C> = true;
+			})
+		constexpr bool single_pass<C> = false;
 
 		// A RandomAccess cursor whose reference_t is a reference type and
 		// whose member type "contiguous" looks like true_type will generate
@@ -165,8 +165,8 @@ STL2_OPEN_NAMESPACE {
 
 		template<class C, class M>
 		META_CONCEPT _Cursor =
-			semiregular<C> &&
-			semiregular<M> &&
+			movable<C> &&
+			movable<M> &&
 			constructible_from<M, C> &&
 			constructible_from<M, const C&>;
 
@@ -289,7 +289,9 @@ STL2_OPEN_NAMESPACE {
 			readable<C> && Next<C>;
 		template<class C>
 		META_CONCEPT Forward =
-			Input<C> && !single_pass<C> && sentinel_for<C, C>;
+			Input<C> && !single_pass<C> &&
+			copyable<C> && copyable<mixin_t<C>> &&
+			constructible_from<mixin_t<C>, const C&> && sentinel_for<C, C>;
 		template<class C>
 		META_CONCEPT Bidirectional =
 			Forward<C> && Prev<C>;
--- a/include/stl2/detail/iterator/counted_iterator.hpp
+++ b/include/stl2/detail/iterator/counted_iterator.hpp
@@ -90,9 +90,13 @@ STL2_OPEN_NAMESPACE {
 			length_ = __counted_iterator::access::count(i);
 			return *this;
 		}
-		constexpr I base() const
+		constexpr I base() const&
 		noexcept(std::is_nothrow_copy_constructible_v<I>) { // strengthened
 			return current_;
+		}
+		constexpr I base() &&
+		noexcept(std::is_nothrow_move_constructible_v<I>) { // strengthened
+			return std::move(current_);
 		}
 		constexpr iter_difference_t<I> count() const noexcept {
 			return length_;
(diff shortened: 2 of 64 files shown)


include/stl2/detail/iterator/basic_iterator.hpp
```diff
@@ -160,10 +160,10 @@ STL2_OPEN_NAMESPACE {
 
 		template<class C, class M>
 		META_CONCEPT _Cursor =
-			Semiregular<C> &&
-			Semiregular<M> &&
+			Movable<C> &&
+			Movable<M> &&
 			Constructible<M, C> &&
-			Constructible<M, const C&>;
+			Constructible<M, C&&>;
```
cjdb: I think this case is handled by the line above it.

include/stl2/detail/iterator/counted_iterator.hpp
```diff
@@ -296,18 +296,34 @@ STL2_OPEN_NAMESPACE {
...
 		template<Iterator I>
+		requires Copyable<I>
+		constexpr auto recounted(const I&, const I& i, iter_difference_t<I> = 0)
+		noexcept(std::is_nothrow_copy_constructible<I>::value)
+		{
+			return i;
+		}
+
+		template<Iterator I>
+		requires !Copyable<I>
```
cjdb: This requirement shouldn't be necessary: the above overload implies that this one happens when `I` doesn't model `Copyable`.

test/algorithm/copy_n.cpp
```diff
@@ -12,21 +12,26 @@
...
 
+//void test() requires <IndirectlyCopyable<I, O>>{
+
+//}
+
 int main() {
-	static const int source[] = {5,4,3,2,1,0};
-	static constexpr std::ptrdiff_t n = sizeof(source)/sizeof(source[0]);
+	single_pass_array source = {5,4,3,2,1,0};
+	static constexpr auto n = std::size(source);
```
cjdb: s/`std`/`ranges`/

test/algorithm/copy_n.cpp
```diff
@@ -12,21 +12,26 @@
...
 	int target[n];
 
 	std::fill_n(target, n, 0);
 	static_assert(n >= 2);
-	auto res = ranges::copy_n(source, n - 2, target);
-	CHECK(res.in == source + n - 2);
+	auto res = ranges::copy_n(source.begin(), n - 2, target);
+	CHECK(&(*res.in) == source.data() + n - 2);
```
cjdb: I know @CaseyCarter gets uncomfortable when seeing `&` used instead of `std::addressof`. I don't know if this is only in parameterised code, or if it's a general thing.

test/single_pass_array.hpp
```diff
@@ -0,0 +1,98 @@
...
+		bool operator!=(const single_pass_pointer_iterator::sentinel & o) const {
+			return false;
+		};
+
+		bool operator==(const single_pass_pointer_iterator & o) const {
+			return !o._p || (o._p == _end );
+		};
+		bool operator!=(const single_pass_pointer_iterator & o) const {
+			return !(*this == o);
+		};
```
cjdb: Let's make these hidden friends.

cjdb: Also, there are trailing semicolons at the end of each function.

test/single_pass_array.hpp
```diff
@@ -0,0 +1,98 @@
...
+	const T & operator*() const {return *_p; };
+	T & operator*() {return *_p; };
+
+
+	bool operator==(const single_pass_pointer_iterator::sentinel& s) const {
+		return s == *this;
+	};
+	bool operator !=(const single_pass_pointer_iterator::sentinel &s) const {
+		return s != *this;
+	};
```
cjdb: Hidden friends, semicolons.

include/stl2/detail/iterator/counted_iterator.hpp
```diff
@@ -296,18 +296,34 @@ STL2_OPEN_NAMESPACE {
...
 	namespace ext {
 		template<Iterator I>
+		requires Copyable<I>
 		constexpr auto uncounted(const I& i)
 		noexcept(std::is_nothrow_copy_constructible<I>::value) {
 			return i;
 		}
 
+		template<Iterator I>
+		requires !Copyable<I>
```
cjdb: This requirement shouldn't be necessary: the above overload implies that this one happens when I doesn't model Copyable.

cjdb (changes requested):
A few comments, overall LGTM.

test/algorithm/copy_n.cpp
CaseyCarter: It's about ADL-hijacking. `&x` can result in a call to some random overload of `operator&` when `x` has program-defined type. In this case, we know that `*res.in` is an lvalue `int`, so there's no problem using `&`.

CaseyCarter:
I rebased this onto master, refactored `ext::uncounted`/`ext::recounted` to require rvalues, and tweaked some things. Let me know / point me at a branch if I've created conflicts for you, and I'll rebase your work onto this. (I have a bad habit of breaking things, but I don't mind cleaning up my own mess.)

> I still have to write tests for 35 or so more algorithms accepting InputIterator.

Don't forget the algorithms that accept `OutputIterator` / `WeaklyIncrementable`. We've allowed *all* single-pass iterators to be move-only, not just `InputIterator`s.

cor3ntin:
@CaseyCarter Thanks a lot - i feel bad... I'll try to make some progress this week. Yes, i was counting all relevant algorithms, including a few OutputIterators (in fill, fill_n and generate, I may miss a few)

PR #284, opened 2019-04-08
alg.set_intersection6:

CaseyCarter:
* Missing `<array>` include
* Why are we comparing a `bool` with `0`?


xiangfan-ms:
There are also three comparisons against 0 in set_intersection.hpp.

CaseyCarter:
> There are also three comparisons against 0 in set_intersection.hpp.

That should be all of them: I audited all occurrences of `lexicographical_compare`.

PR #285, opened 2019-04-09
temporary_vector: temporary_buffer::size_ is inaccessible

CaseyCarter:
Also, rename `_clear` to something less offensive while I'm here.


xiangfan-ms:
Similar issue in include\stl2\detail\algorithm\stable_sort.hpp:

if (!buf.size_) {

PR #291, opened 2019-04-10
Update meta.cpp from upstream

CaseyCarter:


CaseyCarter:
GCC 7.3 seems not to like something in `meta.cpp`, although it compiles fine for me locally. @cjdb, is it time to update https://hub.docker.com/r/cjdb/cmcstl2-testbed? Or is it time for me to learn to make my own docker images? ;)

cjdb:
> CaseyCarter commented 15 days ago

I really need to fix my email filtering (everything GitHub goes into a dark corner).

cjdb:
On Ubuntu Bionic:
* GCC 7.3 times out
* GCC 8.2 passes

On Ubuntu Disco:
* GCC 7.4 times out
* GCC 8.3 ICEs

@CaseyCarter which distro are you using GCC with, and is it a custom build or a package manager install?

PR #309, opened 2019-06-14
Implemented changes from P1716

tomaszkam:
The changes in this papers weakness the constrains of compare
algorithm to be derived from the predicate (required invocation
only in the specific direction).
The IndirectRelation concept is now replaced with IndirectBinaryPredicate,
that requires only single order of argument.
The IndirectlyComparable concept is now derived from IndirectBinaryPredicate.

Updated the following algorithms, to use IndirectRelation:
* find,
* adjacent_find,
* count,
* replace and replace_copy,
* remove and remove_copy,
* unique and unique_copy.

In addition the following two algorithms were changed to use
IndirectlyComparable as they compare elements of two different sequences:
* find_first_of,
* mismatch.


tomaszkam:
This implementation was based on old revision.

PR #310, opened 2019-06-14
Implemented changes from P1716

tomaszkam:
The changes in this papers weakness the constrains of compare
algorithm to be derived from the predicate (required invocation
only in the specific direction).
The IndirectRelation concept is now replaced with IndirectBinaryPredicate,
that requires only single order of argument.
The IndirectlyComparable concept is now derived from IndirectBinaryPredicate.

Updated the following algorithms, to use IndirectRelation:
* find,
* adjacent_find,
* count,
* replace and replace_copy,
* remove and remove_copy,
* unique and unique_copy.

In addition the following two algorithms were changed to use
IndirectlyComparable as they compare elements of two different sequences:
* find_first_of,
* mismatch.

### DIFF
--- a/include/stl2/detail/algorithm/adjacent_find.hpp
+++ b/include/stl2/detail/algorithm/adjacent_find.hpp
@@ -23,7 +23,7 @@
 STL2_OPEN_NAMESPACE {
 	struct __adjacent_find_fn : private __niebloid {
 		template<ForwardIterator I, Sentinel<I> S, class Proj = identity,
-			IndirectRelation<projected<I, Proj>> Pred = equal_to>
+			IndirectBinaryPredicate<projected<I, Proj>, projected<I, Proj>> Pred = equal_to>
 		constexpr I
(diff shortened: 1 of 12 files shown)


tomaszkam:
Confirmed that all affected algorithms are still well-formed, as they are calling `invoke(invoke(proj1, first1), invoke(proj2, first2))`. This includes once that are constrained with `IndirectlyComparable`:

The only not-obvious case is `is_permutation` that counts number of "equal" element in the same range, however it remains valid, because `find_if` and `count_if` are used with following predicate (see https://github.com/CaseyCarter/cmcstl2/blob/master/include/stl2/detail/algorithm/is_permutation.hpp#L188-L190):
```c++
auto match_predicate = [&pred, &e](auto&& x) {
	return __stl2::invoke(pred, e, static_cast<decltype(x)&&>(x));
};
```

tomaszkam:
Added Equivalence and IndirectEquivalence relation that are now used to constrain the following algorithms:
 * is_permutation
 * unique and unique_copy

PR #312, opened 2019-06-22
Numeric ranges

cjdb:
This is an attempt at getting the design for numeric ranges going (third time lucky?). There's a lot of design documentation in the commit messages; and possibly some rambling too.

The only algorithm this PR commits is `accumulate` at present.


include/stl2/detail/functional/arithmetic/minus.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+STL2_OPEN_NAMESPACE {
+	namespace ext {
+		template<class T, class U>
+		META_CONCEPT __differenceable_with = __summable_with<T, U> &&
+			requires(__uncvref<T> t, __uncvref<U> u) {
+				{ +std::forward<T>(t) } -> Same<T>;
+				{ -std::forward<T>(t) } -> Same<T>;
+				{ +std::forward<U>(u) } -> Same<U>;
+				{ -std::forward<U>(u) } -> Same<U>;
```
cjdb: I _think_ these are supposed to be `Common`, not `Same`. I should add tests for the arithmetic operations.

cjdb: Yep, test for `partial_sum` reveals that this is the case.

include/stl2/detail/functional/arithmetic/minus.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <stl2/detail/fwd.hpp>
+#include <stl2/detail/numeric_traits/identity.hpp>
+#include <type_traits>
+#include <utility>
+
+STL2_OPEN_NAMESPACE {
+	namespace ext {
+		template<class T, class U>
+		META_CONCEPT __differenceable_with = __summable_with<T, U> &&
+			requires(__uncvref<T> t, __uncvref<U> u) {
```
cjdb: I don't like that `T` and `U` have had their cv- and ref-qualifiers removed, but I can't think of a better way to implement this at the moment.

include/stl2/detail/functional/arithmetic/minus.hpp
```diff
@@ -0,0 +1,63 @@
...
+				{ +std::forward<T>(t) } -> Same<T>;
+				{ -std::forward<T>(t) } -> Same<T>;
+				{ +std::forward<U>(u) } -> Same<U>;
```
cjdb: Unary `+` possibly considered harmful.

cjdb: Unary `+` removed.

include/stl2/detail/functional/arithmetic/minus.hpp
cjdb: Fixed in cd3992c3f27f90797128f90764839b610f5fe088, which also adds `negate`.

include/stl2/detail/functional/arithmetic/modulus.hpp
```diff
@@ -0,0 +1,64 @@
...
+			requires Magma<modulus, T, U>
+			: t_{t}
+			{}
+
+			constexpr auto value() const {
+				return t_ + T{1};
+			}
+		private:
+			const T& t_;
+		};
```
cjdb: This is not an identity element.

Issue #311, opened 2019-06-21
GCC 8.3 and 9.3: Internal compiler error in type_dependent_expression_p, at cp/pt.c:25930


mnieper:
Running `make` produces the following error with GCC 8.3 and GCC 9.1 on Ubuntu 19.04:

```
[ 11%] Building CXX object test/iterator/CMakeFiles/iter.operations.dir/operations.cpp.o
In file included from /home/mnieper/github/cmcstl2/include/stl2/detail/range/concepts.hpp:24,
                 from /home/mnieper/github/cmcstl2/include/stl2/detail/iterator/insert_iterators.hpp:21,
                 from /home/mnieper/github/cmcstl2/include/stl2/iterator.hpp:28,
                 from /home/mnieper/github/cmcstl2/include/stl2/view/iota.hpp:16,
                 from /home/mnieper/github/cmcstl2/test/iterator/operations.cpp:13:
/home/mnieper/github/cmcstl2/include/stl2/detail/range/access.hpp: In substitution of ‘template<class R> using __begin_t = decltype (std::experimental::ranges::v1::__cpos::begin(declval<R>())) [with R = std::experimental::ranges::v1::ext::take_exactly_view<std::experimental::ranges::v1::iota_view<int> >&]’:
/home/mnieper/github/cmcstl2/include/stl2/detail/range/concepts.hpp:42:8:   required by substitution of ‘template<class T> using iterator_t = std::experimental::ranges::v1::__begin_t<R&> [with T = std::experimental::ranges::v1::ext::take_exactly_view<std::experimental::ranges::v1::iota_view<int> >]’
/home/mnieper/github/cmcstl2/include/stl2/detail/range/concepts.hpp:138:28:   required from here
/home/mnieper/github/cmcstl2/include/stl2/detail/range/access.hpp:53:19: internal compiler error: in type_dependent_expression_p, at cp/pt.c:25930
   53 |     { __decay_copy(r.begin()) } -> Iterator;
      |       ~~~~~~~~~~~~^~~~~~~~~~~
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-9/README.Bugs> for instructions.
make[2]: *** [test/iterator/CMakeFiles/iter.operations.dir/build.make:63: test/iterator/CMakeFiles/iter.operations.dir/operations.cpp.o] Fehler 1
[...]

PR #313, opened 2019-06-24
sets the required standard to C++20

cjdb:
CMake was compiling with -std=gnu++1z by default, which apparently causes some strange functionality (see https://godbolt.org/z/g6duRE). This commit sets the minimum standard to the closest standard to either C++2a or C++20 and disables compiler extensions.


CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ cmake_minimum_required(VERSION 3.8)
 
 project(cmcstl2 CXX)
 
+set(CMAKE_CXX_STANDARD 20)
```
CaseyCarter: This `CMakeLists.txt` eschews "global" properties and attaches properties to targets. We should bump `cxx_std_17` on line 37 to `cxx_std_20` instead.

CMakeLists.txt
```diff
@@ -14,6 +14,9 @@ cmake_minimum_required(VERSION 3.8)
...
+set(CMAKE_CXX_EXTENSIONS OFF)
```
CaseyCarter: This should *not* be a property of the exported `stl2` interface library: we don't want to mandate whether or not users can use compiler extensions. Ideally it would be a property of an internal-only interface library to which all the examples/tests link.

cjdb: Huh, I didn't realise these propagate to users as well.

cjdb: [Looking at this thread][1], I don't know if it's possible to achieve what you're hoping for, at least in CMake 3.8. An alternative would be to change line 37 to `target_compile_options(stl2 INTERFACE -std=c++17)`.

[1]

CaseyCarter: Some investigation demonstrates that the value of `CXX_EXTENSIONS` doesn't propagate into the properties of the installed interface library. Which I suppose is only logical, since there's no way to set the property on an interface library.

CaseyCarter:
```
CMake Error in examples/CMakeLists.txt:
  The compiler feature "cxx_std_20" is not known to CXX compiler

  "GNU"

  version 7.3.0.
```
While we don't precisely *need* any particular 20 features given support for TS concepts, I'm inclined to cut the cord anyway and drop support for GCC7. "Most recent two compiler releases" is reasonable support for a research tool like cmcstl2.

Opinions?

cjdb:
> While we don't precisely need any particular 20 features given support for TS concepts, I'm inclined to cut the cord anyway and drop support for GCC7. "Most recent two compiler releases" is reasonable support for a research tool like cmcstl2.
>
> Opinions?

I'm in favour, but I think we'll need to fix #311 first.

CaseyCarter:
> I'm in favour, but I think we'll need to fix #311 first.

Actually supporting GCC9 is obviously a prerequisite for GCC9 support; there's no need to make any changes (except to the CI scripts) to drop GCC7 support.

PR #315, opened 2019-08-03
closes a corner case in ranges::advance

cjdb:
ranges::advance(i, -n, i) decrements i while `bool(i != i)`, but it was
accidentally doing one too many decrements thanks to the post-checked
loop. This commit changes the loop to a pre-check.

### DIFF
--- a/include/stl2/detail/iterator/operations.hpp
+++ b/include/stl2/detail/iterator/operations.hpp
@@ -29,12 +29,12 @@ STL2_OPEN_NAMESPACE {
 				i += n;
 			} else {
 				if constexpr (BidirectionalIterator<I>) {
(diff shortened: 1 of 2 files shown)


include/stl2/detail/iterator/operations.hpp
```diff
@@ -82,10 +82,10 @@ STL2_OPEN_NAMESPACE {
 			} else {
 				if constexpr (BidirectionalIterator<I> && Same<I, S>) {
 					if (0 > n) {
-						do {
+						while (n != 0 && i != bound) {
```
CaseyCarter: Inconsistent the ordering of `n != 0` is. Un-Yoda the other comparisons in this function we should as well.

PR #327, opened 2019-08-18
adds stride_view to cmcstl2

cjdb:
stride_view is one of the things I regret not getting into C++23, so much so, that I'm planning on:

1. implementing it at CppCon on stage, so people don't use `filter_view` as an ad-hoc `stride`;
2. writing a proposal for it for Belfast (even if it's not discussed till Bulgaria)


include/stl2/view/stride.hpp
```diff
@@ -0,0 +1,283 @@
...
+		private:
+			using parent_t = __maybe_const<is_const, stride_view>;
+			using base_t = __maybe_const<is_const, R>;
+
+
+			friend __iterator<!is_const>;
+
+			parent_t* parent_ = nullptr;
+			iterator_t<base_t> current_{};
+			range_difference_t<base_t> step_{};
```
cjdb: This should be a `non_propagating_cache`, but I've been having difficulty getting it to hold its value.

include/stl2/view/stride.hpp
```diff
@@ -0,0 +1,283 @@
...
+			, current_(std::move(end))
+			, step_{step}
+			{}
+
+			constexpr __iterator(__iterator<!is_const> const& other)
+			requires is_const && convertible_to<iterator_t<R>, iterator_t<base_t>>
+			: parent_(other.parent_)
+			, current_(other.current_)
+			{
+				STL2_EXPECT(true);
```
cjdb: Breakpoint statement needs to go.

include/stl2/view/stride.hpp
```diff
@@ -0,0 +1,283 @@
...
+				}
+
+				ranges::advance(current_, n * parent_->stride_, ranges::end(parent_->base_));
+				return *this;
+			}
+		};
+	} // namespace ext
+
+	namespace views::ext {
+		struct __stride_fn : detail::__pipeable<__stride_fn> {
```
cjdb: Apparently it's not necessary to derive from `detail::__pipeable`.

PR #335, opened 2019-09-04
pulls C++20 names out of ext::

cjdb:
Most of the contents of [P1035][1] were nested inside namespace `__stl2::ext`. Now that P1035 has been merged, this commit pulls those names out into `__stl2` (except for `istream_view`, which will be done after NB comments).

[1]: https://wg21.link/p1035


test/CMakeLists.txt
```diff
@@ -38,3 +38,4 @@ add_subdirectory(iterator)
 add_subdirectory(algorithm)
 add_subdirectory(view)
 add_subdirectory(memory)
+add_subdirectory(strides)
```
CaseyCarter: Leakage.

test/CMakeLists.txt
```diff
@@ -38,4 +38,3 @@ add_subdirectory(iterator)
 add_subdirectory(algorithm)
 add_subdirectory(view)
 add_subdirectory(memory)
-add_subdirectory(strides)
```
cjdb: 🤦‍♂

PR #336, opened 2019-09-10
Changes unreachable to unreachable_sentinel_t, same for default_sentinel

cjdb:
I have a feeling this will be necessary come Monday.

### DIFF
--- a/include/stl2/detail/iterator/unreachable.hpp
+++ b/include/stl2/detail/iterator/unreachable.hpp
@@ -17,28 +17,30 @@
 #include <stl2/detail/iterator/increment.hpp>
 
 ///////////////////////////////////////////////////////////////////////////
-// unreachable [unreachable.sentinel]
+// unreachable_sentinel_t [unreachable_sentinel_t.sentinel]
 //
 STL2_OPEN_NAMESPACE {
-	struct unreachable {
+	struct unreachable_sentinel_t {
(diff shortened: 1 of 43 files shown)


include/stl2/detail/iterator/unreachable.hpp
```diff
@@ -17,28 +17,30 @@
 #include <stl2/detail/iterator/increment.hpp>
 
 ///////////////////////////////////////////////////////////////////////////
-// unreachable [unreachable.sentinel]
+// unreachable_sentinel_t [unreachable_sentinel_t.sentinel]
```
CaseyCarter:
```suggestion
// unreachable_sentinel_t [unreachable.sentinel]
```

test/iterator/counted_iterator.cpp
```diff
@@ -86,23 +86,23 @@ static_assert(test_constexpr());
 int main()
 {
 	using ranges::counted_iterator, ranges::common_iterator;
-	using ranges::default_sentinel, ranges::sized_sentinel_for;
+	using ranges::default_sentinel_t, ranges::default_sentinel, ranges::sized_sentinel_for;
```
CaseyCarter:
```suggestion
	using ranges::default_sentinel, ranges::default_sentinel_t, ranges::sized_sentinel_for;
```

PR #338, opened 2019-09-11
ranges::swap should allow customization for unions

CaseyCarter:
.... which `is_class_v` rejects. Thanks to Jonathan Wakely for catching the bug.

### DIFF
--- a/include/stl2/detail/swap.hpp
+++ b/include/stl2/detail/swap.hpp
@@ -36,6 +36,14 @@ STL2_OPEN_NAMESPACE {
 		return tmp;
 	}
 
+	namespace detail {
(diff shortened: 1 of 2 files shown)


include/stl2/detail/swap.hpp
```diff
@@ -36,6 +36,14 @@ STL2_OPEN_NAMESPACE {
 		return tmp;
 	}
 
+	namespace detail {
+		template<class T>
+		META_CONCEPT has_class_or_enum_type =
+			std::is_class_v<__uncvref<T>> ||
```
CaseyCarter: Use intrinsics here.

PR #340, opened 2019-09-11
Fix regression in ranges::swap on MSVC

CaseyCarter:
... introduced in d392f97.


include/stl2/detail/swap.hpp
```diff
@@ -66,6 +66,10 @@ STL2_OPEN_NAMESPACE {
 		// https://github.com/ericniebler/stl2/issues/139)
 		template<class T> void swap(T&, T&) = delete;
 
+#ifdef STL2_WORKAROUND_MSVC_895622
```
CaseyCarter: `#if`; it's always defined.
```suggestion
#if STL2_WORKAROUND_MSVC_895622
```

PR #346, opened 2019-10-10
add const_view

dvirtz:
I plan to make a proposal out of this and would like to have an implementation looked at beforehand.

Note that the adapter is named `views::as_const` as opposed to `const_` in range-v3. This is inspired by  the fact that `views::move` is conceptually applying `std::move` on every element the same way that this view conceptually applies `std::as_const`. 
Another possible name is `const_elements`, suggested by @cjdb.


include/stl2/view/const.hpp
```diff
@@ -0,0 +1,276 @@
...
+		template <bool>
+		struct __iterator;
+
+	public:
+		const_view() = default;
+		constexpr explicit const_view(R base) : base_(std::move(base)) {}
+
+		constexpr auto base() const noexcept -> R { return base_; }
+
+		constexpr __iterator<false> begin()
```
cjdb: You'll need to adda requirement for  `detail::simple_view` on each of the non-const overloads.

include/stl2/view/const.hpp
```diff
@@ -0,0 +1,276 @@
...
+		}
+		constexpr auto end() const requires range<const R>
+		{
+			return end_impl(*this);
+		}
+
+		constexpr auto size() const requires sized_range<R>
+		{
+			return __stl2::size(base_);
+		}
```
cjdb: A non-`const` overload is necessary for `size`.

include/stl2/view/const.hpp
```diff
@@ -0,0 +1,276 @@
...
+		iterator_t<base_t> current_{};
+
+	public:
+		using iterator_category = iterator_category_t<iterator_t<base_t>>;
+		using value_type = std::add_const_t<range_value_t<base_t>>;
+		using difference_type = range_difference_t<base_t>;
+
+		__iterator() = default;
+
+		constexpr explicit __iterator(parent_t &parent)
```
cjdb: s/`parent &`/`parent& `/

include/stl2/view/const.hpp
```diff
@@ -0,0 +1,276 @@
...
+
+		constexpr __iterator &operator++()
+		{
+			++current_;
+			return *this;
+		}
+
+		constexpr void operator++(int)
+		{
+			current_++;
```
cjdb: Should delegate to `++*this`

include/stl2/view/const.hpp
```diff
@@ -0,0 +1,276 @@
...
+			return x.current_ - y;
+		}
+
+		friend constexpr difference_type operator-(const sentinel_t<base_t> &x, const __iterator &y) requires sized_sentinel_for<iterator_t<base_t>, sentinel_t<base_t>>
+		{
+			return x - y.current_;
+		}
+	};
+
+	template <class R>
```
cjdb: Please add the requirements here too, as well as `viewable_range`.

include/stl2/view/const.hpp
```diff
@@ -0,0 +1,276 @@
...
+	const_view(R &&r)->const_view<all_view<R>>;
+
+	} // namespace ext
+
+	namespace views::ext
+	{
+	struct __as_const_fn : detail::__pipeable<__as_const_fn>
```
cjdb: Inheritance not needed for this one. Also, please look at some of the other adaptors to see what's usually done. I recommend `take` or `drop`.

include/stl2/view/const.hpp
cjdb: Sorry, I meant `!simple_view`.

include/stl2/view/const.hpp
dvirtz: I think it's necessary for adaptors with no arguments. same as `views::move`.

include/stl2/view/const.hpp
```diff
@@ -0,0 +1,276 @@
...
+
+STL2_OPEN_NAMESPACE
+{
+
+	namespace ext
+	{
+	template <input_range R>
+	requires view<R> 
+	class const_view : public view_interface<const_view<R>>
+	{
```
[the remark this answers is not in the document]
dvirtz: Seems like this project needs a .clang-format file

PR #348, opened 2019-10-14
Disable workarounds for GCC 10

jicama:
GCC pre-10 doesn't need any of these workarounds.

### DIFF
--- a/include/stl2/detail/fwd.hpp
+++ b/include/stl2/detail/fwd.hpp
@@ -27,7 +27,7 @@
 #endif // __clang__
 
 #ifndef STL2_WORKAROUND_GCC_69096
- #if defined(__GNUC__) && __GNUC__ >= 6
+ #if defined(__GNUC__) && __GNUC__ >= 6 && __GNUC__ < 10
   // Return type deduction performed *before* checking constraints.
   // https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69096
   #define STL2_WORKAROUND_GCC_69096 1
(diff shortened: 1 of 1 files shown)


include/stl2/detail/fwd.hpp
```diff
@@ -27,7 +27,7 @@
 #endif // __clang__
 
 #ifndef STL2_WORKAROUND_GCC_69096
- #if defined(__GNUC__) && __GNUC__ >= 6
+ #if defined(__GNUC__) && __GNUC__ >= 6 && __GNUC__ < 10
```
CaseyCarter: We don't support GCC < 7, let's simplify this:
```suggestion
 #if defined(__GNUC__) && __GNUC__ < 10
```

include/stl2/detail/fwd.hpp
```diff
@@ -37,7 +37,7 @@
 #endif
 
 #ifndef STL2_WORKAROUND_GCC_79591
- #if defined(__GNUC__) && __GNUC__ >= 6
+ #if defined(__GNUC__) && __GNUC__ >= 6 && __GNUC__ < 10
```
CaseyCarter: Ditto:
```suggestion
 #if defined(__GNUC__) && __GNUC__ < 10
```

CaseyCarter:
Thanks for the cleanup, Jason. (I'm ecstatic about the branch finally being merged!)

include/stl2/detail/fwd.hpp
CaseyCarter: I'll apply these in a followup commit since "allow maintainers to push to this branch" isn't enabled.

include/stl2/detail/fwd.hpp
CaseyCarter: Ditto followup.

